PR #89, opened 2013-04-10
Live mode update delay

yesbox:
From the config comment:

  // Wait this many milliseconds between evaluating input.
  // In effect this should have similar properties to non_blocking mode but
  // without using threads, save more resources and work in Sublime Text 2.
  //
  // live_instant_interval dictates the minimum time between (near) instant
  // evaluations on new input. When the file first receives input the evaluation
  // is always instant, after which live_instant_interval time must pass before
  // the next input is evaluated instantly again.
  //
  // live_continuous_delay dictates the minimum time between delayed evaluations.
  // If input is continuous (faster than the live_instant_interval interval) then
  // GitGutter will wait this long before evaluating.
  //
  // Has no effect with live_mode disabled or non_blocking enabled.
  // Set live_delay to false to evaluate after each input with no delay.
  "live_delay": true,
  "live_instant_interval": 200,
  "live_continuous_delay": 400

### DIFF
--- a/GitGutter.sublime-settings
+++ b/GitGutter.sublime-settings
@@ -2,7 +2,7 @@
   // Custom path to git binary when not in PATH
   "git_binary": "",
   
-  // Live mode evaluates changes every time file is modified,
+  // Live mode evaluates changes every time file is modified
   // Set false to disable evaluation after each input
   "live_mode": true,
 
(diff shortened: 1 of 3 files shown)


rchl:
IMO this adds too much complexity and settings that maybe 0.001 of users would bother to tweak.
Current master has async invocations so the point of this improvement is made mostly moot now (ST2 still uses sync invocations but I don't think there is any point in actively supporting ST2 anymore anyway).

(And this pull request is few years old so probably the author doesn't have interest in it anymore.)

rchl:
With async invocation I don't think this is worth doing anymore. Also it adds quite a lot of unneeded complexity. If you strongly believe that this should be integrated then feel free to argue :)
Thanks for the work though.

yesbox:
The point of this patch was to reduce the latency when typing with GitGutter with the added benefit of somewhat lowered CPU usage/battery drain. If the lag is gone with ST3 the patch no longer serves a purpose. Thanks for the feedback.

PR #131, opened 2014-01-13
added command for Sublime Text 3 to OSX

keeev:


jisaacks:
@keeev I agree that I do not think it is worth repeating the entire thing for a single char change. Please make this more concisely say that you use a 3 instead of a 2 and I will merge this.

Thanks.

PR #136, opened 2014-01-28
Differentiate between staged and unstaged changes

jisaacks:


mokkabonna:
This would be great, I just made a ustage checkout of selected hunk in the git package. https://github.com/kemayo/sublime-text-git/pull/404

With the gutter having no difference between the two it is a bit hard knowing which one to use.

TheSecEng:
This still being looked at?

deathaxe:
I havn't done much about GitGutter at all as most of its features are part of core ST right now. I thought about a more "lightweight" variant to handle stuff, which is not yet part of the core, but hesitated due to some incompatibilities it would imply.

I played around with solutions about displaying staged/unstaged changes but the required changes are quite huge and I ran into several issues which were not so easy to handle. Abandoned those work in the end.

I personally got used to use Sublime Merge for more sophisticated comparisons, so havn't the need to push that feature into GitGutter. Don't even know how to properly mark it in the gutter as we default to ST's default minidiff if it is available.

PR #203, opened 2014-12-05
Allow changing of the compare_against setting globally, or on a per-project basis

natecavanaugh:
My use case for this is that I have multiple projects around a repo where I have pull requests sent against different branches. This setting would make it ideal so that I can just add the setting in my individual projects without having to remember to do it every time.

Also, I can imagine this being useful for those who maybe use Git Flow and are comparing against a branch such as "develop" or "release".


natecavanaugh:
Hi there, just wanted to bump this pull again, since fixing the last comment back on the 10th. Any luck in getting this merged?
Thanks :)

jisaacks:
So sorry @natecavanaugh, got a new job and been busy ramping up for that. So forgot about this. Thanks for brining it to my attention. I will do my best to get this looked at before Sunday. :)

jisaacks:
I have spent some time looking at this code. I want to get it merged in, but I must apologize I think there is a better way it can be implemented.

I think it makes more sense to have the logic moved from `git_gutter_handler` file to the `view_collection` file. It is convenient in the handler because we are already loading the settings there, but I think it is fine to load the settings once in the view collection via a `plugin_loaded` function.

You can do that by putting the function at the end of the file like so:

```python
def plugin_loaded():
    settings = sublime.load_settings('GitGutter.sublime-settings')
    ViewCollection.compare_against = settings.get('compare_against', 'HEAD')
```

Then the `ViewCollection.get_compare` can be defined like so:

```python
@staticmethod
def get_compare(view):
    compare = ViewCollection.compare_against
    return view.settings().get('git_gutter_compare_against', compare)
```

You would then need to change the [3 places](https://github.com/jisaacks/GitGutter/search?utf8=%E2%9C%93&q=get_compare) that this method is called to pass in the view. In the git_gutter_compare method you will need to grap the view like `self.view = self.window.active_view()`, most of the other methods in that file are doing the same thing. The other 2 places already have the view available at `self.view`

Also I think it makes sense to state in the `GitGutter.sublime-settings` comment; something to the effect that this setting is only for changing the default (which defaults to HEAD) and there is a command for temporarily changing.

----

If you have a specific reason why you think my suggestions are wrong, I am open for discussion. But this feels cleaner to me.

Also, thank you for being patient. :+1:

natecavanaugh:
Heya @jisaacks 
I've gone ahead and moved the code over. I don't have any issues with doing it that way. Just let me know if there's anything else you'd like changed.

Thanks! :)

jisaacks:
It looks great. Please give me a couple days to test it out, if there are no problems I will merge.

Thanks!

jisaacks:
OK, I found one issue - I just fixed it myself though. You can read the [commit message](https://github.com/jisaacks/GitGutter/commit/053af19c955d993e4f4105f8726140cad6759809) if curious.

Thanks

natecavanaugh:
Awesome! Thanks for the fix :)

PR #210, opened 2015-01-26
Hexadecimal view fix/workaround.

KevinLi:


jisaacks:
:shipit: looks good to me.

PR #255, opened 2015-10-08
Support environment variables in git_binary path

richardfrost:
I have an environmental variable that I use to reference my portable apps folder that changes from computer to computer. After taking a look at the code for the [Sublime Text Git plugin](https://github.com/kemayo/sublime-text-git) I noticed it wasn't too bad to support loading environmental variables.

Example config on Windows:
```json
{ "git_command": "%Apps%\\Cmder\\vendor\\msysgit\\bin\\git.exe" }
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -79,6 +79,13 @@ If git is not in your PATH, you may need to set the `git_binary` setting to the 
 }
 ```
 
+You can also use environmental variables in your git_binary path (Windows example):
+```json
+{
+  "git_binary": "%EnvVar%\\git\\bin\\git.exe"
+}
+```
+
 
 #### Per-project Settings
 Sublime Text supports project-specific settings, allowing `live_mode` to be set differently for different repositories.
--- a/git_gutter_handler.py
+++ b/git_gutter_handler.py
@@ -254,11 +254,15 @@ class GitGutterHandler:
 
     def run_command(self, args):
         startupinfo = None
+        # Per http://bugs.python.org/issue8557 shell=True is required to
+        # get $PATH on Windows. Yay portable code.
+        shell = os.name == 'nt'
         if os.name == 'nt':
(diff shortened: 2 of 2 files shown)


jisaacks:
It looks like you are only setting [shell=true](https://github.com/jisaacks/GitGutter/pull/255/files#diff-a8bd557be58e1975071210105c1f4fe2R259) and enabling env vars for windows. Is there a reason for that?

richardfrost:
Thanks for getting to this so fast! When I put this together I only had my Windows machine, which is why I included the example for it. I was thinking that most *nix based systems come with git installed, or are a simple command away from being installed, whereas Windows is not usually quite as simple. According to the link I referenced in my comment, shell=true on a POSIX system actually seems to mess up the loading of environmental variables/path, which is why I tried to avoid it.

I tested on an Ubuntu system today and found that it almost worked, but when Popen runs, it is sending the git command wrapped in single quotes, which prevents variables from being interpreted, and therefore doesn't work.

My new proposed fix is to simply pass `git_binary` through `os.path.expandvars(git_binary)` and saving the result to `self.git_binary_path`. [os.path.expandvars](https://docs.python.org/2/library/os.path.html#os.path.expandvars) will return a string with any environmental variables expanded. This works on both Windows and Linux in my tests. We could add a string check to see if it contains a "$" or "%" character, but I think that this is probably good enough.

Sorry for not checking on other operating systems beforehand. I do not have access to OSX, but I'm hopeful since it works on linux that it would work on OSX. If you would like I can create a new clean pull request with this new change, or improve it if need be.

Thanks!

jisaacks:
According to the docs for [expandvars](https://docs.python.org/2/library/os.path.html#os.path.expandvars), it looks like it should Just Work™

Issue #247, opened 2015-08-20
Temporary files are not removed


borntyping:
The temporary files created to compare the buffer against the file in git are never removed, resulting in `/tmp` being filled with multiple copies of files that have been edited.

jisaacks:
What OS are you using?

borntyping:
Linux Mint 17.2, Sublime Text 3083, Git Gutter 1.2.1.

Wintaru:
This happens to me in Debian Jessie. Very annoying, I don't close Sublime Text or reboot my machine very often and my sysadmin brought it to my attention that there were hundreds of source files littering my /tmp folder.

rchl:
Will mark as dup of #284 as that one has more information.

PR #268, opened 2016-02-18
Add an identifiable prefix to temp files

statik213:
After a long sublime session I find my /tmp directory littered with temp files created by Git Gutter. It's worse on Windows where the %TEMP% directory is persistent. It would be helpful to be able to identify files left behind by GitGutter so they can be manually cleaned up.

Ideally Git Gutter would remove these files when they are not needed. I'll work on that next.

### DIFF
--- a/view_collection.py
+++ b/view_collection.py
@@ -83,7 +83,8 @@ class ViewCollection:
     def git_tmp_file(view):
         key = ViewCollection.get_key(view)
         if not key in ViewCollection.git_files:
-            ViewCollection.git_files[key] = tempfile.NamedTemporaryFile()
+            ViewCollection.git_files[key] = \
+                tempfile.NamedTemporaryFile(prefix='git_gutter_')
             ViewCollection.git_files[key].close()
         return ViewCollection.git_files[key]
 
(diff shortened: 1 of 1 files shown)


jisaacks:
@statik213 Sorry I have not merged this yet. I have not manually tested it, but the code looks good.

Do you still have plans to:

> Git Gutter would remove these files when they are not needed. I'll work on that next.

If so that would address #247 

Let me know, thanks!

statik213:
Hi @jisaacks, I took a look at it a while ago but didn't get a chance to implement a fix yet. I'm not exactly clear on what the life cycle for each item in the dicts in view_collections.py is. I'm not sure if entries ever get removed. Removing the files would be easy when you figure out when items should be removed from that dict.

I'd also suggest changing to using tempfile.mkstemp instead of NamedTemporaryFile for a couple of reasons:

  1. tmpfile.name is used in multiple places and the file re-opened, written and closed. NamedTemporaryFile is intended to be used onetime and the file automatically deleted on close(). mkstemp gives you ownership of deleting the file and just returns a path. I think that's more in line with what we are doing here.
 2.  Re-opening the file after a close() is probably not valid because in a rare race another tmpfile could get created with that same name.

jisaacks:
Thanks @statik213 

> I'd also suggest changing to using tempfile.mkstemp instead of NamedTemporaryFile

Is that something you are willing to work on?

statik213:
Yeah, I can take care of that.

PR #282, opened 2016-06-06
Cleanup temp_file usage

statik213:
Here's the followup for #268. This should improve the templfile usage a little. 
I've been running with it for a few days now and haven't run in to any issues (on Linux). However, I did not get to verify the changes on Windows or OSX.  I wouldn't expect any issues on OSX.

### DIFF
--- a/git_gutter_handler.py
+++ b/git_gutter_handler.py
@@ -78,20 +78,22 @@ class GitGutterHandler:
 
         contents = contents.replace(b'\r\n', b'\n')
         contents = contents.replace(b'\r', b'\n')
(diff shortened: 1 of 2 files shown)


jisaacks:
@statik213 thanks for working on this.

@sigod do you use windows or OSX? if so would you be able to test that this PR doesn't break anything?

sigod:
Yes, I use Windows. I'll check in a few days. Currently I'm too busy with the new project.

sigod:
I'm using it for few days now. No issues so far. (Windows 7/10)

Issue #284, opened 2016-06-18
Temporary files are not deleted


deathaxe:
Today I found thousands of files in my TEMP folder on Windows 10 x64. I found SublimeText 3114 and GitGutter to be the creator of all that files. As soon as I open a tracked file within a git repository, GitGutter creates two temporary files, but they are never deleted.

With a look into code I found GitGutter to make use of python's _tempfile_ object for comparison of the view's content and git.

According to pythons [documentation](https://docs.python.org/3/library/tempfile.html#tempfile.NamedTemporaryFile) the method being used to create the tempory files _tempfile.NamedTemporaryFile_ deletes the created temporary file as soon as it is closed.

The conclusions are:
1. At least on Windows, the function does not work as documented - files are not deleted. Even if I explicitly add the _delete=True_ argument, files are not deleted.
2. If the function would work as documented, GitGutter could not as everytime the temporary file is closed, it would be deleted, what happens after each update of the file content.
3. I guess, you should explicitly call _tempfile.NamedTemporaryFile(delete=False)_ and manually delete all temporary files eigher with on_close event or on exit of SublimeText.

I managed to handle this with `entombedvirus's` fork of GitGutter by adding:

> class GitGutterHandler: 
>     ...
>     def **del**(self):
>         try:
>             os.unlink(self.git_temp_file.name)
>             os.unlink(self.buf_temp_file.name)
>         except:
>             pass

This clears the files as soon as a view is closed.

I am not sure what your intention with the static class ViewCollection is and when to clean files here. Unfortunatelly the above mentioned patch does not work for your GitGutter.

Can you have a look onto this bug, please?

rchl:
I've checked my temp directory and found 52000 git_gutter* files there :)

Soon hopefully the ViewCollection will be gone so we can make nicer fix for this.

rchl:
I have tried this in S3 console and it worked fine though. File was deleted.

```
>>> import tempfile
>>> t=tempfile.NamedTemporaryFile(prefix='aaa_test_')
>>> t.write(b'foo')
3
>>> t
<tempfile._TemporaryFileWrapper object at 0x00000193069A8A90>
>>> t.close()
```

Also tried with `with` in a small "plugin":
```
import sublime
import tempfile

tt = tempfile.NamedTemporaryFile(prefix='aaa_test_')
with tt:
  tt.write(b'foo')
```
In that case file is deleted immediately by the `with` construct so this shouldn't be used in GitGutter if file is to live longer.

[...]

PR #285, opened 2016-06-19
FIX for ISSUE #284: Delete temporary files if a view is closed.

deathaxe:
I managed to find a way to delete temporary files and remove them from the ViewCollection object triggered by the on_close() event. This is a real synchronous event callback, but normally deleting two files, should not take much time.


rchl:
The code has changed quite significantly since this pull request and this problem also affects all platforms now. I have a fix that I will submit for review as soon as #321 is merged.

Thanks for working on this.

Issue #294, opened 2016-08-12
'shutil' used but not imported in git_gutter_handler.py


ericschmidtatwork:
https://github.com/jisaacks/GitGutter/blob/master/git_gutter_handler.py#L284

Just ran into this problem with Sublime Text 3. It crashed when trying to find my git binary because the `shutil` module wasn't imported.

GitGutter used to run fine until it randomly stopped this afternoon, and then I discovered this error, so I don't really know what was going on. But adding the import fixed it.

woble:
Getting the following related error:

Sublime Text 3 Build 3120 (Dev channel)

```
Traceback (most recent call last):
  File "git_gutter_events in D:\Programs\Sublime Text 3\Data\Installed Packages\GitGutter-Edge.sublime-package", line 74, in callback
  File "D:\Programs\Sublime Text 3\Data\Installed Packages\GitGutter-Edge.sublime-package\view_collection.py", line 20, in add
  File "D:\Programs\Sublime Text 3\Data\Installed Packages\GitGutter-Edge.sublime-package\git_gutter_handler.py", line 19, in __init__
  File "D:\Programs\Sublime Text 3\Data\Installed Packages\GitGutter-Edge.sublime-package\git_gutter_handler.py", line 284, in load_settings
NameError: global name 'shutil' is not defined
```

lwojciechowski:
Same problem on Ubuntu 16.04. Sublime 3119 Dev.
```
Traceback (most recent call last):
  File "git_gutter_events in /home/lwojciechowski/.config/sublime-text-3/Installed Packages/GitGutter-Edge.sublime-package", line 74, in callback
  File "view_collection in /home/lwojciechowski/.config/sublime-text-3/Installed Packages/GitGutter-Edge.sublime-package", line 20, in add
  File "git_gutter_handler in /home/lwojciechowski/.config/sublime-text-3/Installed Packages/GitGutter-Edge.sublime-package", line 19, in __init__
  File "git_gutter_handler in /home/lwojciechowski/.config/sublime-text-3/Installed Packages/GitGutter-Edge.sublime-package", line 284, in load_settings
NameError: global name 'shutil' is not defined
```

sigod:
#295

PR #242, opened 2015-08-04
Allowing git_binary to be customized on a per-platform basis.

johnburnett:
On Windows, git isn't always in PATH, requiring you to set git_binary for GitGutter to work correctly.  For users that sync settings across platforms, this allows git_binary to be (optionally) customized on a per-platform basis.

PR #295, opened 2016-08-12
Fix issue #294 with shutil errors.

lwojciechowski:
After #242 there was missing `shutil` dependency in `git_gutter_handler.py`.


jisaacks:
Thanks!

PR #296, opened 2016-08-12
Hover popup for inline diffs

r-stein:
This adds a hover popup for inline diffs, available for Build 3116+.

Demonstration:
demo

Update demonstration mit symbol icons:
hover demo

Update demonstration with new icons:
popup_demo

This also includes a dependency for `mdpopups`, which is required for syntax highlighting.


r-stein:
Do you think it would be better to use icons instead of text:
example

jisaacks:
That looks great! Thanks so much for putting this together! :)

I'll need to test it out and review before I can merge though. @sigod would you mind testing on windows?

r-stein:
I have changed the text to symbols now and hope they are self explaining.

Actually I am using windows. So I can confirm it works there. I will test the backward compatibility with Build 3114 and ST2. Please review and feel free to ask questions.
Especially in the part of parsing the git diff I am not so confident with the structure and it's mainly based on the comment above. In addition I assume the flags ensure there are only `+` and `-` lines in the hunk and only extract the `-` lines for the inline diff.

jisaacks:
@r-stein I do not know what I am doing wrong but I cannot get this to work at all. I added a `print`  as the first line to the `on_hover` method and it is never getting called.  This means the hover callback event is never triggering. I have not used sublime hover events before so I may be doing something wrong. I am on latest `3114` version. I also made sure all the deps got installed. Any ideas?

Also, if possible, I would prefer if you could break the event handler into a thin event handler that can be moved to the __events__ file and then call a command that contains most of the logic if it passes all the prechecks. Does that make sense? Not sure if it would be possible but would make the organization better I think.

Thanks! :)

r-stein:
The Hover event is only available in the dev build, it has been added in Build 3116. However you can easily turn it into a windowcommand, just change the start to:
``` py
class GitGutterInlineDiffCommand(sublime_plugin.WindowCommand):
    def run(self):
        view = sublime.active_window().active_view()
        point = view.sel()[0].b
```
and add a keybinding for `git_gutter_inline_diff`

And I will move the event listener to the events file.

sigod:
Very interesting. I'll try it later.

> Do you think it would be better to use icons instead of text

How about an option for this?

r-stein:
I have added a command just in case. This will open the diff popup in the line of the first selection.
We could add a keybinding, e.g.
``` js
    { "keys": ["ctrl+shift+alt+c", "d"], "command": "git_gutter_diff_popup" },
```

In addition I added a setting to switch between text and icon buttons. We can still easily remove it and only keep icon.

r-stein:
I have also added a arrow-button to jump to the first, previous and next change, which will only be available if that change exists (i.e. disable first change if the popup is already the first change):

jump changes

If someone else has an useful idea, feel free to suggest it.

princemaple:
Wow amazing feature!

Does ST support HTML titles? That would be great to make it clear what the icons do.

r-stein:
ST does not seem to support html titles/tooltips yet, although they are still adding new features.

princemaple:
Hm, then I think the text may still be necessary. I wouldn't be able to tell that it's a copy icon there.

r-stein:
Yeah I don't find a good copy icon in unicode =/. Using texts is big and ugly (even bigger if we use "next change" instead of only "next"):
gg_text_buttons

It might be the best to add a short explanation to the readme. I think after using the icons a few times, they are easy to recognize. In addition copying will give a status message, so one can just try to click on all buttons and see what is happening.

princemaple:
> best to add a short explanation to the readme

that and an update message (release note) will be great :smile: (I'm not a maintainer though)

jisaacks:
This is looking great! I wonder if we should wait to release this until `3116` is released? Thoughts?

r-stein:
@jisaacks only the hover event is limited to `3116+`. The popup itself can opened with a command in `3114` (this is not as cool as the hover event, but is not to bad either). We could add a keybinding or even an entry in the command panel. If people don't use it they won't notice it anyway and if people see it in the README and want to use it, they could trigger it manually.
Hence I don't think we need to wait for the new ST release, but should emphazise, that the hover event is not yet available in the README.

*However I am still in the process of improvement and getting feedback.*
Aside: Do you agree, that it would be a good idea to include a screenshot with a short description of the buttons in the README?

jisaacks:
Yeah, that makes since, I just want to be able to mention this feature in the change-log without the caveat of saying the hover won't work yet. Maybe thats not so bad though as long as they can use it with a keyboard shortcut.

> Aside: Do you agree, that it would be a good idea to include a screenshot with a short description of the buttons in the README?

Yeah that sounds good.

jisaacks:
Alright, so I was able to get the popup to open by adding my own keyboard shortcut. We should go ahead and add that to for all OS's since we are for the other commands. Once I got it open though, it was plain white background (my theme is a dark background) and the code was not colored formatted.

jisaacks:
Also as soon as we can get this working nicely and with an easy way for the user to use it (e.g. keyboard shortcut, command palette.) I will go ahead and cut a release with it and just explain in the release notes how to use it now and that hover will work automatically once sublime adds it. :)

r-stein:
It looks like the popup css has also only added in `3116` https://www.sublimetext.com/3dev. I will add a `GitGutter: Show Diff Popup` command. Which keybinding would you suggest? One could use `"ctrl+shift+alt+c", "d"` or `"ctrl+shift+alt+c", "ctrl+d"`  (with `d` for diff) as this prefix is already used.

jisaacks:
> It looks like the popup css has also only added in 3116 https://www.sublimetext.com/3dev

Thats a bummer. I wonder when it will be released. It has been several months since the current 3114 was released. I guess we can go ahead and release, it's still useful as is and will only get more useful. We can also explain that they can upgrade to dev build if they want to get the full effect asap. Thoughts?

>Which keybinding would you suggest?

`"ctrl+shift+alt+c", "d"` seems fine to me, why would we need the extra `"ctrl+d"`?

> (with d for diff) as this prefix is already used.

Which is already used?

r-stein:
>> (with d for diff) as this prefix is already used.
>Which is already used?

`"ctrl+shift+alt+c"` is already used as prefix for commands.

r-stein:
I just realised, that the syntax highlighting was, because the css support changed in 3119 (and I used new css). I have adapted it and it should now be highlighted in 3114.

jisaacks:
> I have adapted it and it should now be highlighted in 3114.

Sweet. I don't see any commits changing anything for this though. (other the the ones adding the command)

r-stein:
The commit is https://github.com/jisaacks/GitGutter/pull/296/commits/f2f166829bdbcecf8d9cf8e699ae4c3f5c48a07b and as remark for the highlighting: mdpopups has a [setting](http://facelessuser.github.io/sublime-markdown-popups/usage/#mdpopupsuse_sublime_highlighter) to switch between pygments and a integrated ST syntax highlighter. It might also be worth to add that to the README.

jisaacks:
@r-stein I'll take a look at this again tonight. Is there anything else you are planning to add? If not, will you go ahead and resolve any merge conflicts. :)

r-stein:
I would add a screenshort and short button description to the README, i.e.:
# Git Gutter

A sublime text 2/3 plugin to show an icon in the gutter area indicating whether a line has been inserted, modified or deleted.

### Screenshot:

screenshot

### Diff Popup

hover_popup_screenshot


| symbol | meaning if the symbol   |
| ---    | ---                     |
| ×      | close the popup         |
| ⤒      | jump to first change    |
| ↑      | jump to previous change |
| ↓      | jump to next change     |
| ⎘      | copy the content        |
| ⟲      | revert changes          |

...

r-stein:
In addition I am still wondering whether it would be better to remove the optional button text and only stay with icons.

r-stein:
Look here: https://github.com/r-stein/GitGutter/tree/hover_popup for the changed README

sigod:
I see on your screenshot the pop up ignores configuration for drawing white space. Can it be fixed?

r-stein:
@sigod The whole code highlighting is done by `mdpopups` not by me. It also uses pygments as highlighter by default (every user can change this at a global level). So it would be an issue [there](https://github.com/facelessuser/sublime-markdown-popups/issues)

jisaacks:
@r-stein can you make the screenshot some other code that is not related to the popup itself. As it is now, the user might think you are actually trying to get them to pay attention to the code (because it says `show_diff_popup`.) Viewers may be confused by that. I think a more generic code example will get people to focus more on the popup itself instead of the code that happens to be in the screen shot.

Also I think a we should move the popup diff section screenshot. Actually as it is getting more verbose. I think this new order might make sense:

### Main Title
 * description /w main screenshot

### Installation

### Not working?

### Advanced Features

* Diff Popup
* Comparing against different commits/branches/tags
* Jumping Between Changes

### Settings


Does that make sense?

r-stein:
Yes, I can change the screenshot and I agree to the structure.

jisaacks:
Also, @sigod @r-stein, speaking of read me structure. If y'all want to add a **maintainers** section with all our names. I am perfectly fine with that. Up to y'all.

r-stein:
Yes, I agree to such a section or just write it below the title
# GitGutter
by X Y, ...

jisaacks:
> or just write it below the title

I think it makes sense to just have a **Mainteners** section at the bottom. Seems to be the standard that I have seen on most projects. It could lists collaborators (@jisaacks, @sigod, @r-stein) directly with link to [all contributors](https://github.com/jisaacks/GitGutter/graphs/contributors)

jisaacks:
@r-stein was just playing around with this. Its working really well. It would be nice if the color scheme in the popup matched or if the popup would extend to show longer lines but those aren't deal breakers.

So as soon as you make those read me changes I'll go ahead merge this and cut a release.

r-stein:
You can put `"mdpopups.use_sublime_highlighter": true,` into your user settings to get a matching color scheme as in the screenshots.

r-stein:
I have changed the README. You can also make additional changes and finish it for the release.

jisaacks:
Sounds great @r-stein I'll try to get this done tonight. :)

jisaacks:
👍 Thanks so much for all your work on this @r-stein

PR #304, opened 2016-08-20
Check the protected regions in the whole line.

r-stein:
This checks whether the protect region is anywhere in the line and not only at the start of the line. Since the gutter icon is applied to the whole line, it should also checked in the whole line, whether a gutter icon has been added.


jisaacks:
@r-stein I see you are doing a lot of work on GitGutter. Would you like me to add you as a collaborator?

r-stein:
I actually just saw this forum thread https://forum.sublimetext.com/t/setting-priority-for-gutter-icons/8861/5 . However I would like to be added as a collaborator.

Issue #305, opened 2016-08-23
git binary lost after upgrade to 1.2.4


okote:
Hi,

After the automatic upgrade from 1.2.3 to 1.2.4, I got the message:
`Your Git binary cannot be found.  If it is installed, add it to your PATH environment variable, or add a 'git_binary' setting in the 'User/GitGutter.sublime-settings' file.`
which never happened to me with previous versions.

I fixed it following the instructions, but it may be a problem for some users. It may be a symptom of other issues, though. I'm using the default git at `/usr/bin/git` in a macOS Sierra (beta 6), with ST2 2.0.2 (2221).

Cheers,

Nicolás

degamer106:
Same here.  Don't know what changed.  I'll try the instructions as well.  

I'm running Sublime Text 2 + this plugin on OSX 10.11.6 if that helps.

jisaacks:
Prior to the new version if GitGutter could not find your `git` binary it would silently fail. Now it shows this warning.

okote:
But *it was* able to find it before. I've been using GitGutter for months, without problem.

r-stein:
I think the problem is that it uses `shutils.which` on ST3, which is not available in ST2. However one could run a manual search for the executable.

mccarthy:
This is not working as intended.  For me, the git binary is at `/usr/local/bin/git`, and is in my PATH, but I am still seeing this error message.  

I'm on:
ST 2.0.2 Build 2221
OSX 10.10.5
GitGutter 1.2.4

r-stein:
Might be a good idea to merge and release https://github.com/jisaacks/GitGutter/pull/306 as a Bugfix before the next bigger release.

rchl:
@mccarthy It's in the PATH of your terminal but probably not of the Sublime. Since some time OSX is quite restrictive on regarding which environment variables GUI applications inherit. You probably would have to install something like https://github.com/int3h/SublimeFixMacPath to make Sublime find git from `/usr/local/bin/git` automatically.

rchl:
And #306 was merged so I think there is resolved now.

Roman1972:
REMOVE GitGutter And everything goes !!!

PR #306, opened 2016-08-23
Search for git executable in the PATH for ST2

r-stein:
Fix for https://github.com/jisaacks/GitGutter/issues/305
This adds a search for the git executable on ST2. It also adds a caching for previously found executable and should slightly speed up ST3.

### DIFF
--- a/git_gutter_handler.py
+++ b/git_gutter_handler.py
@@ -18,6 +18,7 @@ except (ImportError, ValueError):
 
 class GitGutterHandler:
     git_binary_path_error_shown = False
+    git_binary_path_fallback = None
 
     def __init__(self, view):
         self.load_settings()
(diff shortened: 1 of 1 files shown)


git_gutter_handler.py
```diff
@@ -284,8 +285,19 @@ def load_settings(self):
 
         if self.git_binary_path:
             self.git_binary_path = os.path.expandvars(self.git_binary_path)
+        elif self.git_binary_path_fallback:
+            self.git_binary_path = self.git_binary_path_fallback
```
jisaacks: should this be `GitGutterHandler. git_binary_path_fallback` instead of `self.git_binary_path_fallback`, I could be wrong, I didn't fully comprehend this yet, just glanced at it.

r-stein: when retrieving `self.git_binary_path_fallback` will use the non-static field if present and the static otherwise and `GitGutterHandler.git_binary_path_fallback` will always use the static field. In the current code it would not make any difference, because we don't have a non-static field.

jisaacks: **TIL**  Did not know that _instance_ members would fallback to _static_ members if unset. I don't do much python outside of GitGutter, never seen that in another language before.

jisaacks:
:shipit: LGTM

r-stein:
I have only tested it on windows with ST2 and ST3

jisaacks:
@r-stein thanks, I will test this on OSX with ST3 (when I get a chance) If it breaks linux/ST2 I will rely on people reporting issues. I have yet to find an integration tool for ST plugins and I don't want to get into the business of manually testing every version on every OS :)

PR #309, opened 2016-08-29
Mode to highlight the differences inside the diff popup

r-stein:
Add a mode to toggle the highlight mode of the diff popup between "syntax highlighting" and "differences highlighting". The "differences highlighting" will show the merged content of the git and the editor and highlight differences with different styles.

diff_highlight

Current styling:

| color | meaning |
| --- | --- |
| red | text was _deleted_ |
| yellow | text was deleted and _replaced_ by new text |
| green | text was _inserted_ and might replace deleted text |

_I am still looking forward to do some changes, but it is ready for test and feedback._


jlegewie:
I like this addition!! It would be great if it supported "Beautiful diffs" from git 2.9 (see [here](https://github.com/blog/2188-git-2-9-has-been-released)). 

I would also prefer if the diff works with a transparent background color (like [here](https://github.com/blog/2188-git-2-9-has-been-released)) and not with the font color. Not sure how 
easy that is.

Finally, there should be an option to make this the default.

r-stein:
> I like this addition!! It would be great if it supported "Beautiful diffs" from git 2.9 (see here).

I am currently using [difflib](https://docs.python.org/3.3/library/difflib.html) to compare the diff, which is generated to show the gutter marks. So I don't see special support for that. However enabling the feature in your git config might improve both the gutter marks and the diff popup.

> I would also prefer if the diff works with a transparent background color (like here) and not with the font color. Not sure how
> easy that is.

I am looking forward add a class for each type (equal, inserted, deleted, replace-inserted, replace-deleted) instead of inline styles and it will be configurable via a css file. Though ST minihtml css is very limited.

> Finally, there should be an option to make this the default.

I was looking forward to add a setting for that (the keybinding has an argument already)

rchl:
Should also update documentation.

r-stein:
@jlegewie now there is a css file to configure each diff style

@rchl I updatet the settings section, but I am not sure how much I should describe the new functionality in the README

git_gutter_popup.py
```diff
@@ -79,6 +82,15 @@ def navigate(href):
...
             view.hide_popup()
             view.run_command("git_gutter")
+        elif href in ["disable_hl_diff", "enable_hl_diff"]:
+            do_diff = {
+                "disable_hl_diff": False,
+                "enable_hl_diff": True
+            }.get(href)
+            # show a diff popup with the same diff info
+            _show_diff_popup_impl(
+                view, point, highlight_diff=do_diff, flags=0,
```
rchl: I think there is no code that currently specifies `flags` but it would still be a correct thing to do to pass originally passed flags here (`flags=flags`).

git_gutter_popup.py
```diff
@@ -89,7 +101,9 @@ def navigate(href):
 
             def show_new_popup():
                 if view.visible_region().contains(pt):
-                    show_diff_popup(view, pt, flags=flags)
+                    show_diff_popup(
+                        view, pt, highlight_diff=highlight_diff,
+                        flags=0)
```
rchl: Shouldn't it be `flags=flags` like before?

git_gutter_popup.py
```diff
@@ -115,35 +131,57 @@ def is_button_enabled(k):
...
+            .format(button)
+        )
+
+    if highlight_diff:
+        # (*) show a highlighted diff of the merged git and editor content
+        min_indent = _get_min_indent(lines + meta["added_lines"])
 
-    if not is_added:
+        git_content = "\n".join(l[min_indent:] for l in lines)
+        editor_content = "\n".join(l[min_indent:] for l in meta["added_lines"])
```
rchl: Nit: Maybe better to call these "old_content" and "new_content"? Or "previous_content", "current_content"?
That is because I don't think "git_content" is very accurate name when it only really includes removed lines.

git_gutter_popup.py
```diff
@@ -152,27 +190,112 @@ def is_button_enabled(k):
...
+    css = sublime.load_resource("Packages/GitGutter/gitgutter_popup.css")
+    try:
+        user_css = sublime.load_resource("Packages/User/gitgutter_popup.css")
+        css += "\n"
+        css += user_css
+    except OSError:
+        pass
+
+    # apply the jinja template
+    jinja_kwargs = {
```
rchl: This object is not used.

git_gutter_popup.py
```diff
@@ -152,27 +190,112 @@ def is_button_enabled(k):
...
+def _to_html(s):
+    return (
+        html.escape(s, quote=False)
+        .replace("\n", "<br>")
+        .replace(" ", "&nbsp;")
+        .replace("\u00A0", "&nbsp;")
+    )
+
+
+def _tag_open(style):
```
rchl: Nit: With such generic name maybe it should have a tag_name argument also?

gitgutter_popup.css
```diff
@@ -0,0 +1,40 @@
+/*
+    Jinja2 css template. Provided values are:
+        - wrapper_class: the surrounding class;
+    Copy the values you want to change into your user directory
+    and change them there to overwrite the stylesheet.
+*/
+
+{{wrapper_class}} .gitgutter-button {
+    font-size: 15px;
```
rchl: I'd love for this to be bigger (maybe 18px). Currently it requires so much precision that it makes my wrist hurt almost. :)
Of course you might choose to do that change in a separate fix.

git_gutter_popup.py
```diff
@@ -1,5 +1,6 @@
+import difflib
+import html
 from functools import partial
-
```
r-stein: Actually I didn't want to remove that line break. I will add it back.

git_gutter_popup.py
```diff
@@ -1,5 +1,6 @@
+import difflib
+import html
```
r-stein: I will take a look at that.

git_gutter_popup.py
r-stein: I played around with this and thought this would be better. The normal flag is "HIDE_MOUSE_ON_MOVE_AWAY". However if someone actively clicked on the popup he mostlikely want it to be persistant, so I make it persistant.

git_gutter_popup.py
r-stein: Since the style it not inline anymore I will remove it anyway.

gitgutter_popup.css
r-stein: I think I will switch to rem, because it is recommended in [minihtml](https://www.sublimetext.com/docs/3/minihtml.html) and maybe also make it a little bit bigger =)

git_gutter_popup.py
r-stein: Can you be a little more specific about this? I don't see an error on ST2 (it doesn't work there, because the popup is ST3 only).

rchl: But it throws there because there is no such module in ST2. At least on Windows.
Would suggest moving it to where mdpopups is imported.
Possibly same for difflib as you don't need it in ST2.

r-stein: I don't have the problem on windows with ST2, but sounds fair to only import them when necessary.

git_gutter_popup.py
```diff
@@ -1,24 +1,22 @@
-import difflib
-import html
-from functools import partial
-
 import sublime
 import sublime_plugin
 
 try:
+    if int(sublime.version()) < 3080:
+        raise Exception("No popup available.")
```
rchl: This exception will be caught so I don't really see the point of raising it. Or at least for having text for it.

r-stein: Yeah it is just to avoid everything to be imported, since it is not necessary on ST2 anyway. I want to check for version 3080, because it introduces the popup, even if mdpopups is available.
Yes, the message is not important at all.

git_gutter_popup.py
rchl: The intention might be told clearer by explicitly removing that bitfield from the mask. Although in practice it won't really matter. At least current set of flags available and used.

git_gutter_popup.py
```diff
@@ -79,17 +81,28 @@ def navigate(href):
...
             sublime.status_message("Copied: " + copy_message)
         elif href in ["next_change", "prev_change", "first_change"]:
             next_line = meta.get(href, line)
             pt = view.text_point(next_line - 1, 0)
 
             def show_new_popup():
                 if view.visible_region().contains(pt):
-                    show_diff_popup(view, pt, flags=flags)
+                    show_diff_popup(
+                        view, pt, highlight_diff=highlight_diff,
```
rchl: That was an old problem but git_gutter_handler argument is missing here. But instead of adding that, you should probably call *_impl() here.

rchl: Actually this is supposed to show a new diff so this needs to be solved differently. I'll add the fix to #325 and you should probably make the same change here.

r-stein: The reason to call show_diff_popup (unsted of _impl as above) again is, that the point and therefore the line has changes. The diff_info is for that specific line/hunk and we must update that.

rchl: Yes, I've added a fix in #325. As I've said there, I don't care too much which pull request integrates those.

rchl: Not sure I understand what you mean.
This code is used for navigating between changed chunks in the file. Content of the file doesn't change while doing that.

r-stein: Okay then I misunderstood something and will apply your suggestion

git_gutter_popup.py
```diff
@@ -100,9 +100,12 @@ def navigate(href):
...
             def show_new_popup():
                 if view.visible_region().contains(pt):
-                    show_diff_popup(
-                        view, pt, highlight_diff=highlight_diff,
-                        flags=0)
+                    popup_kwargs = {
+                        'action': 'show_diff_popup',
+                        'point': pt,
+                        'flags': 0
+                    }
```
rchl: You don't want to preservce `highlight_diff` state?

r-stein: ups thanks

jisaacks:
:shipit:

Issue #230, opened 2015-05-17
GitGutter causes massive lag when editing over sshfs


egrefen:
I'm using ST3 with v1.1.2 of GitGutter to edit files on an sshfs mount on OS X 10.10.3.

GitGutter is causing massive lag (one or two seconds delay after a character is typed before it appears, spinning beach ball when switching tabs), in that normal behaviour resumes when GitGutter is disabled.

I tried setting `"non_blocking": true` in the GitGutter user config, but this did not help. I tried playing with various combinations of the other options, but to no avail. Only disabling GitGutter seems to work, when working remotely. Is there anything I should have tried that I haven't, or are other people encountering this too?

jisaacks:
This is a known issue. Git gutter is running git commands as you make changes to determine the icons to show. When you are editing a file over a network it will be having to run git commands over that network and will be very slow. The current answer is to not use git gutter when editing files over a network.

I would be open to ideas or PRs that attempt to solve this issue but it is not an issue I felt the need to try to solve since I think the majority use something like `vim` when editing files over a network.

Related: #79

DecodedMonkey:
I quick fix would be to add an option to turn off the constant checking and have it only check when a key cmd is pressed. This would allow for people to run the check once they have written some code.

UltraNurd:
I run into a similar issue when working on a file on a share mounted over SMB over VPN. (I cannot log directly into the storage appliance to run an editor, though I could get a shell on a closer machine.) It's definitely git that is slow (the commands can take a very long time when I run them in Terminal myself), but I'm not sure why it's locking up the Sublime Text UI when non_blocking is true.

adityavm:
:+1: This is still an issue. If `non_blocking: true` worked as intended, it would go a long way towards fixing this I feel.

ddarbyson:
Any updates on this issue? I'm mounting  SFTP drives via ExpanDrive 5 and having lag issues.

rchl:
Please re-open for now. That rewrite doesn't introduce async invocations yet. That will come next.

jisaacks:
Oops. That was an auto closing from the commit message I believe.

rchl:
More advanced users should feel free to test the fix for this in #321.

rchl:
#321 integrated. Should be resolved in versions newer than 1.3.0.

PR #240, opened 2015-07-20
Async git invocations

entombedvirus:
Disclaimer: This pull req contains extensive changes to GitGutter and I can totally understand if you would rather not accept it.

That said, I use GitGutter regularly on large-ish git repos (~8GB) and I've grown so accustomed to it that  having had to turn GitGutter off due to noticeable delays while typing made me miss it sorely. I tried many combinations of non-blocking, live-mode & debounce_delay and still could not achieve the kind of user experience I wanted: for GitGutter not to introduce _any_ typing lag regardless of size of the git repo.

Some initial experimentation led to the conclusion that calling subprocess.Popen(...) in a separate thread using sublime.set_timeout_async gets rid of the typing lag. But that meant changing the codebase from blocking calls to some kind of async callback style code. I chose to use a poor man's implementation of Promise API for async callbacks.

These are the main highlights of this pull request:
1. git is invoked on a separate thread as to not introduce any delay while typing
2. the compare target is now per git dir. Previously, if you were comparing against tag/0.1.1 in one git repo, GitGutter would diff against that tag on every single window regardless of whether they were separate git projects. Now it will only diff against tag/0.1.1 for projects that share the same .git dir
3. Fixed "Compare against origin" so that it's functional

**I have only tested this on Sublime Text 3 running on OS X as that's the only platform I currently have access to.**

PR #315, opened 2016-10-04
Add GitGutterSettings class that encapsulates settings (#230 related)

rchl:
This is first step towards rewrite of the code to have less globals and more
passing of objects around.

Created GitGutterSettings class which is created in ViewCollection and passed
onto GitGutterHandler's created from there.

The goal is to remove the ViewCollection and pass settings directly from the
base *Command class. Currently settings object is still created in a couple of
places (ViewCollection, individual "WindowCommand"s and EventListener). In the
end it will be created only from the base TextCommand and EventListener.

### DIFF
--- a/git_gutter_settings.py
+++ b/git_gutter_settings.py
@@ -19,11 +19,6 @@ class GitGutterSettings:
         self._settings = None
         self._user_settings = None
         self._git_binary_path_fallback = None
-        # a mapping of git_dir to a string indicating what to compare against.
-        # ex:
-        #   /Users/foo/workspace1/.git -> "origin",
-        #   /Users/foo/workspace2/.git -> "HEAD"
-        self._compare_against_mapping = {}
         # These settings have public getters as they go through more
         # complex initialization than just getting the value from settings.
         self.git_binary_path = None
@@ -106,14 +101,6 @@ class GitGutterSettings:
             self._settings.get('show_status')
         if self.show_status != 'all' and self.show_status != 'none':
             self.show_status = 'default'
-
-    def get_compare_against(self, git_dir):
-        if git_dir in self._compare_against_mapping:
-            return self._compare_against_mapping[git_dir]
-        return self.get('git_gutter_compare_against', 'HEAD')
-
-    def set_compare_against(self, git_dir, new_compare_against):
-        self._compare_against_mapping[git_dir] = new_compare_against
 
 settings = GitGutterSettings()
 
--- a/git_gutter_settings.py
+++ b/git_gutter_settings.py
@@ -43,8 +43,9 @@ class GitGutterSettings:
             'Preferences.sublime-settings')
 
         # Git Binary Setting
-        git_binary_setting = self._user_settings.get("git_binary") or \
-            self._settings.get("git_binary")
+        git_binary_setting = (
+            self._user_settings.get("git_binary") or
+            self._settings.get("git_binary"))
         if isinstance(git_binary_setting, dict):
             self.git_binary_path = git_binary_setting.get(sublime.platform())
             if not self.git_binary_path:
@@ -93,12 +94,14 @@ class GitGutterSettings:
             self.patience_switch = '--patience'
 
         # Show in minimap
-        self.show_in_minimap = self._user_settings.get('show_in_minimap') or \
-            self._settings.get('show_in_minimap')
+        self.show_in_minimap = (
+            self._user_settings.get('show_in_minimap') or
+            self._settings.get('show_in_minimap'))
 
         # Show information in status bar
-        self.show_status = self._user_settings.get('show_status') or \
-            self._settings.get('show_status')
+        self.show_status = (
+            self._user_settings.get('show_status') or
+            self._settings.get('show_status'))
         if self.show_status != 'all' and self.show_status != 'none':
             self.show_status = 'default'
 
--- a/git_gutter_settings.py
+++ b/git_gutter_settings.py
@@ -1,0 +1,109 @@
+import shutil
+import sublime
+
+ST3 = int(sublime.version()) >= 3006
+
+
+class GitGutterSettings:
+    git_binary_path_error_shown = False
+    git_binary_path_fallback = None
+
+    def __init__(self):
+        self._settings = None
+        self._user_settings = None
+        self._git_binary_path_fallback = None
+        # a mapping of git_dir to a string indicating what to compare against.
+        # ex:
+        #   /Users/foo/workspace1/.git -> "origin",
+        #   /Users/foo/workspace2/.git -> "HEAD"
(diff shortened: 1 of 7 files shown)


jisaacks:
So far looks good to me. I have not looked at the code thoroughly tho or tested locally. @sigod @r-stein do either you have time to look this over or test?

r-stein:
1. The code works on my Win7, but I don't really see the point in moving the settings from a global variable to a class attribute/field. The main reason to avoid globals is to avoid global state changes with side effects, but settings are a global state anyway and we don't change them. Hence from my point of view this is not a reasonable goal.
2. A settings wrapper to preprocess things like git path may still be useful.
3. We should avoid calling `self.settings.load_settings()`, which calls `sublime.load_settings`,  in the `__init__` because
  > At importing time, plugins may not call any API functions, with the exception of `sublime.version()`, `sublime.platform()`, `sublime.architecture()` and `sublime.channel()`.
  [API Reference](https://www.sublimetext.com/docs/3/api_reference.html)

rchl:
@r-stein
Ad. 1.
You probably should admit that even if it doesn't make sense for the reasons you provided, it makes sense for a couple other reasons. Currently settings are loaded separately in basically all files (git_gutter.py, git_gutter_change.py, git_gutter_events.py, git_gutter_handler.py, git_gutter_popup.py, view_collection.py). It makes sense to have one class that handles settings for those reasons:
 - *Testability*. Plugin doesn't have tests currently so it's a moot point, but doesn't hurt to make it easier to test.
 - *Readability and maintainability*. Better to have same API everywhere than repeating code in all files and potentially having to modify all places if some logic/names change.
 - *Handling setting changes*. Currently if we would want to add change listener for settings (which makes sense but it's a separate issue) it would have to be done separately in all places.

Now, as I'm basing these changes on original work from #240, which has one instance of settings per view, I was going for the same in the end. I see that this is probably unnecessary, though, so could modify this approach to have minimal number of settings instances, but there would be at least two with the way packages architecture is designed in sublime, I believe.

So would you be fine with solution that keeps settings in their own class but makes them be a global instance?

Ad. 4
I shall fix that.

r-stein:
Although I appreciate your enthusiasm, I don't really understand your points.

- Why would it be easier to test? If I change a setting in one setting object it will be propagated to all settings object. So I could easily change the settings for test purposes for any solution.
- Neither the Sublime API nor the settings name will change so I don't see a maintainability problem. Readability may be debatable, but I don't think something one can significantly increase the readability of `settings.get("key", default_value)`.
- How would you implement the change listener?

Also the number of settings instances is not relevant. Each instance just contains the settings id and a little object overhead and takes a negligible to create. Often plugins create a setting object each time they read a setting value.

The wrapper class would only be necessary to preprocess the setting for the command line arguments in `git_gutter_handler.py`?

rchl:
> Why would it be easier to test? 

Because then settings would be mockable. Globals stand in a way of that.

> Neither the Sublime API nor the settings name will change so I don't see a maintainability problem. 

As long as you are repeating code in all the files, maintainably suffers.
https://en.wikipedia.org/wiki/Don%27t_repeat_yourself

> How would you implement the change listener?

`sublime.Settings.add_on_change` would either reload settings or just update specific keys and notify all listeners (various parts of the code would have to register a listener). In case of GitGutter it seems like most settings are checked directly every time they are needed to be checked so it doesn't apply to most of them, but there are some that are only initialized once (like git_binary) and those could be updated.

Anyways, are you really suggesting that I should keep settings loading separate in all files and just have wrapper for some settings? Wouldn't that be less consistent? And as mentioned before, harder to maintain?

It seems like I have troubles convincing you that this change is better, so to avoid wasting more time, please state exactly what you would rather see or just say that this won't be accepted.

rchl:
I've added commit that creates only one instance of settings as global. See if you like it more.

r-stein:
Yes, I like it more with this change.

Just to add, that settings have globally the same value see (from `sublime.py`):

``` py
class Settings(object):
    def __init__(self, id):
        self.settings_id = id

    def get(self, key, default=None):
        if default is not None:
            return sublime_api.settings_get_default(self.settings_id, key, default)
        else:
            return sublime_api.settings_get(self.settings_id, key)
    # ...
```
So the result of locals, globals or instance variables is the same as long as they have the same `id` (`GitGutter.sublime-settings`).

ldionne:
My understanding is that this PR was one step towards applying the giant PR proposed in #240, which is a good thing in itself.

git_gutter_settings.py
```diff
@@ -0,0 +1,108 @@
+import os
+import shutil
+import sublime
+
+ST3 = int(sublime.version()) >= 3000
+
+settings = None
```
r-stein: You initialize settings here and in line 106, so you could remove this.

git_gutter_settings.py
```diff
@@ -0,0 +1,108 @@
...
+        else:
+            self.ignore_whitespace = ''
+
+        # Patience Setting
+        self.patience_switch = ''
+        if self._settings.get('patience'):
+            self.patience_switch = '--patience'
+
+        # Show in minimap
+        self.show_in_minimap = self._user_settings.get('show_in_minimap') or \
```
r-stein: Also I am aware that this is not from you, it might be better to use parentheses for line breaks as recommended in [pep8](https://www.python.org/dev/peps/pep-0008/#maximum-line-length):

> The preferred way of wrapping long lines is by using Python's implied line continuation inside parentheses, brackets and braces. Long lines can be broken over multiple lines by wrapping expressions in parentheses. These should be used in preference to using a backslash for line continuation.

jisaacks:
@r-stein thanks for looking this over and approving. What did you test this on Windows? Also @rchl what did you test this on? I can try to test this on OSX tonight.

r-stein:
@jisaacks Yes, it works on WIN with ST2 and ST3. (Also there are no OS dependent changes)

git_gutter_settings.py
rchl: Having settings initialized here servers the purpose of self-documenting the code that it's an exposed global. Otherwise it wouldn't be clear IMO.

git_gutter_settings.py
rchl: Updated those three places to use parenthesis.

rchl:
Thanks for merging.

BTW. I intended for the changes to be squashed before integration. I didn't intend for the fixup commits to end up in the final branch history. It seems like Github has a way of [squashing all commits before merging](https://help.github.com/articles/merging-a-pull-request/#merging-a-pull-request-using-the-github-web-interface) so that would be a good option to use next time. :)

jisaacks:
@rchl thanks for pointing that out. This whole time I thought other people were cloning, manually squashing and then merging. That makes things simple!

PR #318, opened 2016-10-15
WindowCommand to TextCommand rewrite

### LINKED ISSUE #230: GitGutter causes massive lag when editing over sshfs

rchl:
Big rewrite based on #240 to prepare for introduction of async invocations to fix #230


r-stein:
It is a nice idea to pass the branch to compare as an argument, but we should keep the other commands (`git_gutter`, `git_gutter_next_change`, `git_gutter_prev_change`, and `git_gutter_compare_against`).
*It might be better to pass `forward` argument to `git_gutter_next_change` instead of the `git_gutter_prev_change` command, but this would break existing user created keybindings.*

rchl:
`git_gutter` command without arguments still works as before.

> but we should keep the other commands (git_gutter, git_gutter_next_change, git_gutter_prev_change, and git_gutter_compare_against)

The problem with having them separate is that there will be no way of getting the existing GitGutterHandler from the main GitGutter command. At least not in a clean way.

What I could probably do is to still have those commands you ask for but those would just act as a proxy and just invoke the newly introduced command on the view (`view.run_command('git_gutter', args=...`).

r-stein:
I would suggest the proxy solution for `git_gutter_next_change` and `git_gutter_prev_change`, because I assume several people have created a keybinding for those.

jisaacks:
Do these changes affect performance in anyway?

rchl:
I would assume not, but I wouldn't know how to verify that really.

The main difference is that, I assume, Sublime internally creates one instance of TextCommand for each view instead of like before where it only created one instance of WindowCommand per window. But that alone seems like a negligible overhead, if you ask me.

rchl:
I have placed a print in GitGutterCommand constructor and noticed that it's instantiated more times than necessary for each view.
```
<GitGutter.git_gutter.GitGutterCommand object at 0x00000000050B6B00> <sublime.View object at 0x00000000049EDC18>
<GitGutter.git_gutter.GitGutterCommand object at 0x00000000050B6E80> <sublime.View object at 0x00000000049EDC18>
<GitGutter.git_gutter_popup.GitGutterDiffPopupCommand object at 0x00000000050B65F8> <sublime.View object at 0x00000000049EDC18>
```
Two are expected, not three. It seems that importing `from git_gutter import GitGutterCommand` import makes this happen. Seems like git_gutter.py is evaluated twice in this case. I think I would also need to merge `GitGutterDiffPopupCommand` into main command to get rid of it.

rchl:
Addressed that in latest commit.

rchl:
Any update? I'm running with this code for my daily work and haven't spotted any problems.

rchl:
Guys? Anyone has time to have a look and review?

jisaacks:
@rchl can you explain the overall goal of this PR. It seems you have removed the ViewCollection in favor of passing around the git_handler to individual classes. All of which are not sub commands on a single git gutter command. Other than reorganization if this buying us anything?

I am not pushing back, just trying to understand if the goal is purely reorganization for future optimizations or if this has some other benefit?

rchl:
ViewCollection was a hack in the first place. It was a global instance managing git handlers per view where the proper solution is to just have instances of TextCommand class per view. That is what TextCommand is designed for. Not using WindowCommand and then adding views managing on top of that. That really didn't make sense from the design point of view. Using WindowCommand didn't have any advantage. Those are to be used when the command relates to the window rather than individual view. And in case of GitGutter commands work in context of the view.

And in general, I think I shouldn't need to convince anyone that globals are bad. Try adding some testing while having a global. Especially in Sublime where tests run in the same scope and at the same time when normal plugin code does.

So all in all this is to clean up code, make it easier to maintain, less likely to break and most importantly, to prepare for the next change which will introduce async invocations. Maybe it would be possible to add async handling on top of old ViewCollection but to be fair I haven't investigated that myself. These changes are based on the work done by someone else and I just personally thought they are very nice,  make the code much nicer and easier to work with.

I understand you might be worried about such big changes introducing regressions but not doing them, on the other hand, makes it more difficult to fix problems and halts progress IMO.

rchl:
Unittests are supported through this package for example: https://github.com/randy3k/UnitTesting
https://github.com/niosus/EasyClangComplete has some tests that you can look at as an example.
EasyClangComplete package even integrates its testing using github so that those run for each new commit (including pull requests).

r-stein:
I am not sure about this changes, although I think that we can get rid of ViewCollection, because it has a lot of redundancy. However unifying all commands into one and dispatching them based on the arguments doesn't seem to be nice either.
For the unit tests: if we would change the key from view.file_name() to view.id() we would also have a `GitGutterHandler` per view. So the usual test "create a view .. do stuff .. assert result/correct regions" would also work.

Nonetheless I think this is acceptable and since all commands can be proxied it won't have disadvantages for the user and running async events will be a huge improvement.

git_gutter.py
```diff
@@ -1,157 +1,64 @@
...
-                    'deleted_dual', 'inserted', 'changed',
-                    'untracked', 'ignored']
-
-    def run(self, force_refresh=False):
-        self.view = self.window.active_view()
-        if not self.view:
-            # View is not ready yet, try again later.
-            sublime.set_timeout(self.run, 1)
+    from git_gutter_handler import GitGutterHandler
+    from git_gutter_compare import GitGutterCompareCommit, \
```
r-stein: Please also wrap long import statements with parentheses

rchl:
As much as I would hate to delay integration of this change even more, I'm willing to do some rewrites to separate "Change" commands into its own command.

The idea I have is to separate parts needed by the "Change" commands from GitGutterHandler into separate class. "Change" commands don't really need most of the code from there, they can trigger some actions in GitGutterHandler through run_command('git_gutter') call with arguments.

rchl:
Added commits that revert to using original commands as that was the controversial part.
We still need to make them act as proxy commands as only the main command creates the GitGutterHandler instance.

rchl:
Any progress with testing?

jisaacks:
I am very sorry @rchl, I have not forgot about you! I had a really busy weekend, forgot about it being Halloween. I am sure there won't be any issues but even still I want to discuss this PR with @r-stein since he has been doing a lot of maintaining lately and seemed to have some concerns.

@r-stein what are your thoughts on this now? I know you had some concerns earlier, have they been addressed?

@r-stein, @sigod also, do you have a twitter where I can I ask you some (unrelated to this PR) questions, if not, would y'all be up for a gitgutter slack channel, where we can discuss stuff?

git_gutter_compare.py
```diff
@@ -1,52 +1,56 @@
...
 
-class GitGutterCompareCommit(sublime_plugin.WindowCommand):
-    def run(self):
-        self.view = self.window.active_view()
-        self.handler = ViewCollection.get_handler(self.view)
 
-        self.results = self.commit_list()
-        if self.results:
-            self.window.show_quick_panel(self.results, self.on_select)
+class GitGutterCompareCommit():
```
r-stein: Please inherit from `object` (`GitGutterCompareCommit(object)`). On python 3 this is redundant, but for python 2 there are minor differences.

git_gutter_handler.py
```diff
@@ -2,29 +2,55 @@
...
     from . import git_helper
     from .git_gutter_settings import settings
-    from .view_collection import ViewCollection
 except (ImportError, ValueError):
     import git_helper
     from git_gutter_settings import settings
-    from view_collection import ViewCollection
 
 
 class GitGutterHandler:
```
r-stein: Same (inherit `object`)

git_gutter_jump_to_changes.py
```diff
@@ -2,13 +2,15 @@
...
 try:
     from .git_gutter_settings import settings
-    from .view_collection import ViewCollection
 except (ImportError, ValueError):
     from git_gutter_settings import settings
-    from view_collection import ViewCollection
 
 
-class GitGutterBaseChangeCommand(sublime_plugin.WindowCommand):
+class GitGutterJumpToChanges:
```
r-stein: same (inherit `object`)

git_gutter_show_diff.py
```diff
@@ -0,0 +1,158 @@
...
+
+try:
+    from .git_gutter_settings import settings
+except (ImportError, ValueError):
+    from git_gutter_settings import settings
+
+ST3 = int(sublime.version()) >= 3000
+
+
+class GitGutterShowDiff:
```
r-stein: same (inherit `object`)

r-stein:
@jisaacks I still don't like combining all commands into one to much. However since I don't have a better idea, I don't see a real issue with it, and it will lead to a major improvement, I approve to merge the PR.

I don't have twitter or slack, but you can reach my via the email adress in my github account.

rchl:
Addressed latest comments and cleaned up branch history.

jisaacks:
@r-stein great thanks :)

@rchl I will merge this soon. There is another PR you are planning after this right? We could probably cut another release after we get them all merged.

rchl:
Yes, next one is planned.

PR #320, opened 2016-11-03
Don't make spaces unbreakable with the new mdpopups

r-stein:
mdpopups converts spaces into non-breakable spaces on its own in version 1.11.0 https://github.com/facelessuser/sublime-markdown-popups/issues/14, so this conversion is not necessary if that version is used.


jisaacks:
Is there a way we can just specify in the dependencies that we need version 1.11.0 or higher?

r-stein:
I don't think so. Dependencies update automatically or can manually updated via the `Package Control: Satisfy Dependencies` command. Also this it not such an big issue, the old version has no problems, but it is just unnecessary.

PR #321, opened 2016-11-08
Introduce async invocations of git (real non-blocking behavior) (#230)

rchl:
GitGutterHandler.run_command() now runs git in an async thread and returns a
Promise which gets resolved when git operation finishes. For clarity of the
code, rather than using hard to read callbacks, added simple implementation of
Promise.

The behavior is always non-blocking now (only in ST3), so removed "non_blocking"
setting.

### DIFF
--- a/git_gutter_compare.py
+++ b/git_gutter_compare.py
@@ -1,10 +1,13 @@
-import sublime
 from functools import partial
+
+import sublime
 
 try:
     from .git_gutter_settings import settings
+    from .promise import Promise
 except (ImportError, ValueError):
     from git_gutter_settings import settings
+    from promise import Promise
 
 
 class GitGutterCompareCommit(object):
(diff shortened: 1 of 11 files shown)


jisaacks:
Thanks for putting all the work into this. 

Why did you add a long comment describing what promises are in the top of your Promise class? Also all the comments in the class seem to follow a format like they are going to be parsed by some type of doc generator. Are you planning on something like that? Just curious.

r-stein:
Works on ST2+3 without a problem for me.
Also the comments are quite helpful wenn writing code, e.g. Anaconda shows them in a popup. They are especially important for utilities as the Promise.

rchl:
I was trying to follow proper docstring formatting (roughly https://www.python.org/dev/peps/pep-0257/ )
And I thought many might not be familiar with what Promises are so tried to be as descriptive as possible.
But feel free to point out any details that you think I should do differently.

rchl:
I agree. Let's run with it for a while. I've been doing some last-minute refactorings of Promise code so I'm not 100% sure myself that everything works fine. :)

jisaacks:
> But feel free to point out any details that you think I should do differently.

Nope I was just asking. I personally live in the JavaScript world where promises are well known, but I don't know how well known they are among the Python community. But I wasn't criticizing, just asking.

Again thanks for all your hard work. Once everyone feels confidant enough to merge this in I'll cut a new release.

Also excited to see you added some testing stuff! 💯

rchl:
Been using whole day at work and didn't see a problem. :)
I would like to maybe sneak in one more fix before release. For temporary files creations not being deleted.

git_gutter_compare.py
```diff
@@ -13,19 +14,25 @@ def __init__(self, view, git_handler):
...
+        self.commit_list().then(self._show_quick_panel)
 
     def commit_list(self):
-        result = self.git_handler.git_commits().decode("utf-8")
-        return [r.split('\a', 2) for r in result.strip().split('\n')]
+        def decode_and_parse_commit_list(result):
+            commit_lines = result.decode("utf-8").strip().split('\n')
+            return [r.split('\a', 2) for r in commit_lines]
+        if not self.git_handler.on_disk():
+            return Promise.resolve([])
```
r-stein: It looks like you don't import the Promise.

rchl:
All seems good to me. Everyone else happy too? :)

deathaxe:
> Why did you add a long comment describing what promises are in the top of your Promise class?
I am happy about this brief description as I basically saw constructs like that in js but always asked myself how to efficiently use them in other languages like python. This time I learned how those constructs are named - Promise.

IMO each code should be self explaining. A short description like the Promise docstring about what a class does in general and how to use it should always be a part of code.

PR #323, opened 2016-11-11
Make sure temporary files are deleted after view is closed (fixes #284)

### LINKED ISSUE #284: Temporary files are not deleted

rchl:
Delete temporary files on destruction of GitGutterHandler class. Chosen to
remove them from destructor rather than from on_view_close event as it's more
reliable. During on_view_close event, the git command might still be running so
it might be too early to delete temporary files. This seems to work even when
closing Sublime with many files where git gutter has triggered.

Also:
 - Create GitGutterHandler lazily, when any of the commands is run rather
   than on creating the instance of the GitGutter text command itself. That
   ensures that we only create temporary files when actually needed rather than
   immediately for every view that is open in the editor.
 - Disable GitGutter in all native input fields (for example search field shown
   on pressing ctrl+f). All those views don't need GitGutter for anything.

### DIFF
--- a/git_gutter.py
+++ b/git_gutter.py
@@ -21,10 +21,20 @@ except (ImportError, ValueError):
 class GitGutterCommand(sublime_plugin.TextCommand):
     def __init__(self, *args, **kwargs):
         sublime_plugin.TextCommand.__init__(self, *args, **kwargs)
-        self.git_handler = GitGutterHandler(self.view)
-        self.show_diff_handler = GitGutterShowDiff(self.view, self.git_handler)
+        self.is_valid_view = self.view.settings().get('is_widget') is not True
+        self.git_handler = None
+        self.show_diff_handler = None
+
+    def is_enabled(self, **kwargs):
+        return self.is_valid_view
 
     def run(self, edit, **kwargs):
+        if not self.git_handler:
+            self.git_handler = GitGutterHandler(self.view)
+        if not self.show_diff_handler:
+            self.show_diff_handler = GitGutterShowDiff(
+                self.view, self.git_handler)
+
         if not self.git_handler.on_disk() or not self.git_handler.git_dir:
             return
 
--- a/git_gutter_handler.py
+++ b/git_gutter_handler.py
@@ -35,6 +35,10 @@ class GitGutterHandler(object):
 
         self._last_refresh_time_git_file = 0
 
+    def __del__(self):
+        os.unlink(self.git_temp_file)
+        os.unlink(self.buf_temp_file)
+
     @staticmethod
     def tmp_file():
         '''


deathaxe:
I tested the changes from this pull request and highly recommend merging.

jisaacks:
@rchl want to cut a release ASAP. Is this the last PR before I do that?

rchl:
This and the #325 are the last ones.

rchl:
I'm sorry. #324 also. #324 and #325 are regression fixes for my changes so important to have them.

jisaacks:
@rchl ok so once we merge in #323, #324 and #325. Do you think they are all good to merge now?

jisaacks:
I'd like to get #309 also

rchl:
Yes, I personally think they are all good and safe. #309 is a bigger change but I've tested and reviewed it so feel pretty confident about it too.

r-stein:
From view the code of #309 is ready, but I am not sure how about the styling and what we should include into the README (e.g. an addition screenshot).

jisaacks:
@r-stein I would just add some text to the [Diff Popup](https://github.com/jisaacks/GitGutter#diff-popup) section in the README explaining the behavior. If you want to add the screenshot from the PR to the README thats fine too. However, the README is starting to get really long. I am wondering if maybe we should make the main README only about the main functionality and all other things like (popup, compare against, changing styles, etc.) should go in seperate READMEs that the main read me can link to? That doesn't need to happen for this release but might be good for next release? What do you think?

r-stein:
Using a separate documentation is a good idea. I added small descriptions about the button and the css in the the readme. I don't think it would be useful to add a description of the new feature in plaintext without a screenshot and maybe a table. So I would just let it as it is and hope the term "difference highlighting" together with the colors is clear enough.

jisaacks:
@r-stein thats fine, I'll also mention it in the changelog when I release. Once we break out the README into seperate docs we should go ahead and add the screenshot/table etc.

deathaxe:
May I suggest to create the `buf_temp_file` and `git_temp_file` even more lazily? Instead of creating them in the GitGutterHandler.__init_() I moved them to `update_buf_file()` and `update_git_file()` to see what happens.

The big advantage is, they are not created for files, which are not part of a git repository. 

At the moment temporary files are always created as soon as any file is opened, but they are empty, if the view's file is not part of a git repo.

I could create a pull request based on this one, If you want.

rchl:
I'm fine with improving it later. Let's just release new version with this fix first. :)

deathaxe:
Ok, this release will push GitGutter to a new level anyway. The whole Promise thing is really impressive. No more lags while editing and Sublime Texts. Many people will be happy about it.

PR #325, opened 2016-11-12
Fix unitialized variable being used in diff popup

rchl:


rchl:
These changes should also be on #309 and I don't care too much if they are merged through this or #309.
But it kinda makes sense to merge through here as these are regression fixes not directly related to #309.

jisaacks:
@rchl it looks like there is a conflict now after merging #309. Can you resolve and then I'll rebase/merge.

jisaacks:
@rchl FYI it looks like you might have a different email in your `git config` than what you have for **github**. So all your commits aren't getting tied to your user and I think that explains why you are not showing up in the commit graph. I am not sure if you care or not but just thought I'd point that out.

If you want to fix it, you should just be able to add the email under your `git config --global user.email` as an additional email on your account and I believe all your previous commits should update to your account.

Again, not necessary,  up to you.

r-stein:
This has been included in #309.

jisaacks:
Ahh good catch. @r-stein that explains the conflict then!

rchl:
@jisaacks thanks for making me aware of that, added other email now too. :)

jisaacks:
@rchl no problem :) you're commits are showing up in the [contributors graph](https://github.com/jisaacks/GitGutter/graphs/contributors) now.

OK, I think everything for the next release is merged. So Unless I hear differently I'll plan on cutting a release tomorrow. 👍 

cc @r-stein @sigod

PR #331, opened 2016-11-16
Some usage optimization

deathaxe:
Hello guys, very much thanks for your incredible good work with GitGutter 1.4.0. 

A also made up my mind how to improve this plug-in a little bit and found two things I would like to discuss or already started to in the issues section.

Sorry for I was not yet able to distinguish the commits to make a pull request for each.

1. Create temp files only for views, whose content is stored in a git repository. (trivial minor improvement)
2. Run Git and diff only if necessary and avoid running it twice.

I think it's better to talk about real code, than generally discuss things. I know there are some concerns about the second probosal and I am sure it needs extensive testing before productive use, but I think it could generally help saving some CPU in the background.

The basic idea behind it is to let git_gutter run only for visible views if live_mode or focus_change_mode are active as git_gutter will run for each view being activated anyway.

I was already able to prevent git_gutter from running 100 times when doing a search and replace in files with this commit and did not find significant issues with gutter not being updated for none active but visible views. 

Hope you are willing to test and discuss it. :-)

### DIFF
--- a/git_gutter_events.py
+++ b/git_gutter_events.py
@@ -61,7 +61,8 @@ class GitGutterEvents(sublime_plugin.EventListener):
         live_mode or focus_change_mode enabled as they would trigger
         git_gutter with the next on_activate() event.
         """
-        if not self.live_mode() and not self.focus_change_mode() or self.is_view_visible(view):
+        if self.is_view_visible(view) or not (
+           self.live_mode() or self.focus_change_mode()):
             self.debounce(view, 'post-save')
 
     def on_activated(self, view):
@@ -113,6 +114,10 @@ class GitGutterEvents(sublime_plugin.EventListener):
         return settings.get('focus_change_mode', default)
 
     def is_view_visible(self, view):
+        """Return true if the view is visible.
+
+        Only an active view of a group is visible.
+        """
         w = view.window()
         return any(view == w.active_view_in_group(g)
                    for g in range(w.num_groups())) if w is not None else False
--- a/git_gutter_events.py
+++ b/git_gutter_events.py
@@ -62,7 +62,7 @@ class GitGutterEvents(sublime_plugin.EventListener):
         git_gutter with the next on_activate() event.
         """
         if self.is_view_visible(view) or not (
-           self.live_mode() or self.focus_change_mode()):
+                self.live_mode() or self.focus_change_mode()):
             self.debounce(view, 'post-save')
 
     def on_activated(self, view):
--- a/git_gutter_events.py
+++ b/git_gutter_events.py
@@ -61,7 +61,7 @@ class GitGutterEvents(sublime_plugin.EventListener):
         live_mode or focus_change_mode enabled as they would trigger
         git_gutter with the next on_activate() event.
         """
-        if not self.change_mode() or self.is_view_visible(view):
+        if not self.live_mode() and not self.focus_change_mode() or self.is_view_visible(view):
             self.debounce(view, 'post-save')
 
     def on_activated(self, view):
@@ -70,7 +70,7 @@ class GitGutterEvents(sublime_plugin.EventListener):
         This event is also called after a file is opened into an
         active view.
         """
-        if self.change_mode():
+        if self.live_mode() or self.focus_change_mode():
             self.debounce(view, 'activated')
 
     def on_hover(self, view, point, hover_zone):
@@ -112,15 +112,7 @@ class GitGutterEvents(sublime_plugin.EventListener):
     def focus_change_mode(self, default=True):
         return settings.get('focus_change_mode', default)
 
-    def change_mode(self):
-        return self.live_mode() or self.focus_change_mode()
-
     def is_view_visible(self, view):
-        """Return true if the view is visible.
-
-        Only an active view of a group is visible.
-        Note: this should be part of the View class but it isn't.
-        """
         w = view.window()
         return any(view == w.active_view_in_group(g)
                    for g in range(w.num_groups())) if w is not None else False
--- a/git_gutter_events.py
+++ b/git_gutter_events.py
@@ -40,21 +40,37 @@ class GitGutterEvents(sublime_plugin.EventListener):
     # Synchronous
 
     def on_modified(self, view):
-        if self.live_mode():
+        """Run git_gutter for visible view.
+
+        The on_modified() is called when typing into an active view
+        and might be called for inactive views if the file changes on disk.
+
+        If the view is not visible, git_gutter will be triggered
+        by on_activate() later. So it's useless here.
+        """
+        if self.live_mode() and self.is_view_visible(view):
             self.debounce(view, 'modified')
 
     def on_clone(self, view):
         self.debounce(view, 'clone')
 
     def on_post_save(self, view):
-        self.debounce(view, 'post-save')
+        """Run git_gutter after saving.
 
-    def on_load(self, view):
-        if self.live_mode():
-            self.debounce(view, 'load')
+        If the view is not visible git_gutter does not run with
+        live_mode or focus_change_mode enabled as they would trigger
+        git_gutter with the next on_activate() event.
+        """
+        if not self.change_mode() or self.is_view_visible(view):
+            self.debounce(view, 'post-save')
 
     def on_activated(self, view):
-        if self.focus_change_mode():
+        """Run git_gutter when the view is activated.
+
+        This event is also called after a file is opened into an
+        active view.
+        """
+        if self.change_mode():
             self.debounce(view, 'activated')
 
     def on_hover(self, view, point, hover_zone):
@@ -95,3 +111,16 @@ class GitGutterEvents(sublime_plugin.EventListener):
 
     def focus_change_mode(self, default=True):
         return settings.get('focus_change_mode', default)
+
+    def change_mode(self):
+        return self.live_mode() or self.focus_change_mode()
+
+    def is_view_visible(self, view):
+        """Return true if the view is visible.
+
+        Only an active view of a group is visible.
+        Note: this should be part of the View class but it isn't.
+        """
+        w = view.window()
+        return any(view == w.active_view_in_group(g)
+                   for g in range(w.num_groups())) if w is not None else False
--- a/git_gutter_handler.py
+++ b/git_gutter_handler.py
@@ -81,8 +81,8 @@ class GitGutterHandler(object):
 
     def on_disk(self):
         # if the view is saved to disk
-        fn = self.view.file_name()
-        on_disk = (fn is not None) and os.path.exists(fn)
+        file_name = self.view.file_name()
+        on_disk = file_name is not None and os.path.isfile(file_name)
         if on_disk:
             self.git_tree = self.git_tree or git_helper.git_tree(self.view)
             self.git_dir = self.git_dir or git_helper.git_dir(self.git_tree)
--- a/git_gutter_handler.py
+++ b/git_gutter_handler.py
@@ -323,13 +323,13 @@ class GitGutterHandler(object):
                 If the view's file is not part of the index
                 git returns empty output to stdout.
                 """
-                return bool(results[1])
+                return bool(results)
 
             args = [
                 settings.git_binary_path,
                 '--git-dir=' + self.git_dir,
                 '--work-tree=' + self.git_tree,
-                'ls-files', '--other', '--exclude-standard'
+                'ls-files', '--other', '--exclude-standard',
             ] + additional_args + [
                 os.path.join(self.git_tree, self.git_path),
             ]
--- a/git_gutter_handler.py
+++ b/git_gutter_handler.py
@@ -26,8 +26,8 @@ class GitGutterHandler(object):
     def __init__(self, view):
         self.view = view
 
-        self.git_temp_file = GitGutterHandler.tmp_file()
-        self.buf_temp_file = GitGutterHandler.tmp_file()
+        self.git_temp_file = None
+        self.buf_temp_file = None
 
         self.git_tree = None
         self.git_dir = None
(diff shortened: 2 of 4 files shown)


git_gutter_handler.py
```diff
@@ -120,8 +125,8 @@ def write_file(contents):
                 f.write(contents)
 
         if self.git_time() > self.git_file_update_interval_secs:
-            with open(self.git_temp_file, 'w'):
-                pass
+            if not self.git_temp_file:
+                self.git_temp_file = GitGutterHandler.tmp_file()
```
rchl: Original code truncated the file but maybe that's not necessary. In some rare case when exception is thrown and caught below, it might change behavior, but that would probably fail for other reasons also (promise would not be resolved).

rchl:
I'll be testing this today but so far I have randomly seen a problem with all lines marked as added.

git_gutter_events.py
```diff
@@ -95,3 +111,16 @@ def live_mode(self, default=True):
...
         return settings.get('focus_change_mode', default)
+
+    def change_mode(self):
+        return self.live_mode() or self.focus_change_mode()
+
+    def is_view_visible(self, view):
+        """Return true if the view is visible.
+
+        Only an active view of a group is visible.
+        Note: this should be part of the View class but it isn't.
```
deathaxe: You mean it's not required and can be deleted? This is a hint for you, only.

git_gutter_handler.py
deathaxe: Could change it to if ... else ... to ensure truncation.

deathaxe:
> ... I have randomly seen a problem with all lines marked as added.

Me, too. But need to clearify why. Have one untracked file in a repo, just as `show_markers_on_untracked_file` was true, but isn't. But this means git_gutter ran. Maybe something with tempfile creation.

git_gutter_events.py
```diff
@@ -61,7 +61,8 @@ def on_post_save(self, view):
         live_mode or focus_change_mode enabled as they would trigger
         git_gutter with the next on_activate() event.
         """
-        if not self.live_mode() and not self.focus_change_mode() or self.is_view_visible(view):
+        if self.is_view_visible(view) or not (
+           self.live_mode() or self.focus_change_mode()):
```
rchl: Sorry for nitpicking but this is weird indentation (3 spaces). Should be 8 spaces.

git_gutter_show_diff.py
```diff
@@ -39,20 +41,29 @@ def bind_untracked(is_untracked):
...
                 self.git_handler.ignored().then(bind_ignored_or_untracked)
-            else:
+
+            # show_untracked was set to false recently so clear gutter
+            elif self.show_untracked:
                 self._clear_all()
+                self._update_status(0, 0, 0, "", "")
+
+            self.show_untracked = show_untracked
+
```
rchl: Would personally prefer no blank lines before `else` or `elif`. Makes it look disconnected from whole if block.

rchl:
Readme says that `show_markers_on_untracked_file` is by default on which isn't the case. Maybe you could also update that. Or maybe the setting should be on by default, to match what was happening until your fix.

rchl:
> Readme says that show_markers_on_untracked_file is by default on which isn't the case. Maybe you could also update that. Or maybe the setting should be on by default, to match what was happening until your fix.

Actually, I was confused. Let me see if we are on the same page.

With `show_markers_on_untracked_file` off (default):
 - before your fix: untracked files show `+` marker on each line
 - after your fix: untracked don't show any markers

With `show_markers_on_untracked_file` on:
 - before your fix: untracked files show `*` marker on each line
 - after your fix: untracked files show `*` marker on each line

So I misunderstood that setting and maybe had too. It seems like it enables special marker when on, but when off, we are still supposed to see `+` markers.

I guess @jisaacks should chime in what is exactly the expected behaviour.

git_gutter_handler.py
rchl: I guess it's not a big deal. Probably can stay like that.

git_gutter_handler.py
```diff
@@ -81,7 +81,7 @@ def _get_view_encoding(self):
 
     def on_disk(self):
         # if the view is saved to disk
-        on_disk = self.view.file_name() is not None
+        on_disk = os.path.exists(self.view.file_name())
```
rchl: This will throw if file name is `None`. Which it is in many cases. In any case you should leave the comparison with `None` as it's cheaper check than accessing file system.

deathaxe: This was my concern, too, but the console shows an `Unable to open /C/Apps/Sublime/Data/Packages/User/GitGutter.sublime-settings` only, whose origin is not the os.path.exists() funciton. No other traceback or something like that here. Adding the `is not None` check doesn't change anything.

rchl: Where is that console message coming from. Neither from the GitGutter nor from Python I suppose? Basically I think that's unrelated message.
Just check yourself in the python console:

```
>>> import os
>>> os.path.exists(None)
Traceback (most recent call last):
  File "<string>", line 1, in <module>
  File "./python3.3/genericpath.py", line 18, in exists
TypeError: can't specify None for path argument
```

deathaxe:
> So I misunderstood that setting and maybe you have too. It seems like it enables special marker when on, but when off, we are still supposed to see + markers.

This behaviour is as you recently posted.

With `show_markers_on_untracked_file: false` all untracked, ignored or files not existing on disk are marked with `+`. As the setting name quite clearly says, they shouldn't be there.

If this setting is true, GitGutter shows `x` for ignored and `*` for untracked files. If a file is neither untracked nor ignored `+` is displayed on each line. That's why the last commit with the on_disk() fix.

I would be very unhappy, if I was not able to disable markers on untracked or ignored files at all. And it does not make sense to me to show them for files, which I told git to ignore as this is the same as the file would not be part of the repo.

rchl:
> I would be very unhappy, if I was not able to disable markers on untracked or ignored files at all.

I think the default should be to show the markers to match current behaviour. What I would suggest to do is to take your fix but also update default setting of that option to true to match documentation and current behaviour.

But I'm thinking those changes are crossing the boundaries of "optimizations" and maybe it would be better to split them into separate pull request.

git_gutter_show_diff.py
deathaxe: Remove blank lines before else or elif only or after "endif", too? I think to have read something about blank lines after an indention block in some style guidline - google or so. But I am not quite sure and see people handling this very differntly. For single lined statements, you are right. This looks a bit disconnected, but with larger blocks of code?

git_gutter_handler.py
deathaxe: I was sticking too much to this Side-By-Side settings feature, that I did forget to test with simply creating a new file to see whats happening. Of course you are very right with that :-)

deathaxe:
> and maybe it would be better to split them into separate pull request.

Just to ask, as I did not do several pull requests in parallel right now: I would need to create a new feature-branch for that and move all changes over there based on the latest official commit of jisaacks?

git_gutter_handler.py
```diff
@@ -81,7 +81,8 @@ def _get_view_encoding(self):
 
     def on_disk(self):
         # if the view is saved to disk
-        on_disk = os.path.exists(self.view.file_name())
+        fn = self.view.file_name()
+        on_disk = (fn is not None) and os.path.exists(fn)
```
deathaxe: You mean that?
```python
file_name = self.view.file_name()
on_disk = file_name is not None and os.path.isfile(file_name)
```

git_gutter_events.py
rchl: I'd say it should be either multiplication of indentation used (in this case 4 spaces) or aligned with parenthesis (in other cases). 3 spaces just seemed wrong. :)

git_gutter_show_diff.py
rchl: This looks wrong to me at least:
```
if foo:
  ...

else bar:
  ...
```
The blank line before `else`.

git_gutter_handler.py
rchl: To be fair just ```file_name and os.path.isfile(file_name)``` would be enough. Empty string is probably not gonna happen but it would be equally invalid as `None` if it did. But I don't insist on changing it.

git_gutter_handler.py
```diff
@@ -111,40 +92,11 @@ def update_buf_file(self):
         if not self.buf_temp_file:
             self.buf_temp_file = GitGutterHandler.tmp_file()
 
-        with open(self.buf_temp_file, 'wb') as f:
+        with open(self.buf_temp_file, 'wb', 1024 * 1024) as f:
```
rchl: Is specifying buffer size really necessary? What's the advantage? I think that this will potentially just use 1MB of memory for every open file. I think leaving buffer size for python to determine would be preferable.

deathaxe: This was just a wrong assumtion. I intended to prevent smaller files from being written to disk but kept in file buffer, but this can obviously not work. Will revert that.

git_gutter_handler.py
```diff
@@ -86,13 +75,13 @@ def update_buf_file(self):
...
             contents = self.view.substr(region).encode('utf-8')
 
-        contents = contents.replace(b'\r\n', b'\n')
-        contents = contents.replace(b'\r', b'\n')
-
         if not self.buf_temp_file:
-            self.buf_temp_file = GitGutterHandler.tmp_file()
+            # hopefully create the tempfile with attribute O_TEMPFILE,
+            # so Windows keeps file in RAM instead of scratching disk.
+            fd, self.buf_temp_file = tempfile.mkstemp(prefix='git_gutter_')
```
rchl: a) Can you back up your claim about keeping it in RAM and O_TEMPFILE attribute (btw. I think you meant O_TMPFILE)? Documentation doesn't mention anything like that [1]. Besides, even if this API really does some optimizations to file writing to avoid writing to disk, they will not be in effect because you are closing the file so that particular call must end up with file being written to disk. Otherwise how would other part of the code be able to open it again?

b) You are talking about O_TMPFILE (misspelled as O_TEMPFILE, I believe) and Windows which I think doesn't make sense because O_TMPFILE seems to be unix or linux specific.

c) This code is platform independent so you shouldn't mention Windows specifically in any case.

[1] https://docs.python.org/3/library/tempfile.html#tempfile.mkstemp

deathaxe: OK.

deathaxe: The basic idea / hope is that python's `tempfile` handles temporary files efficient on each platform, but if not it does not make sence to open a file to get its name, close it and then reopen it again. Sure this will happen once only, but would be useless if `tempfile` has no advantages.

I keep using `tempfile` with the hope python will make of platform functions handle tempfiles optimal on each platform. As the file is updated very frequently with `live_mode: true` I just want to avoid real disk usage too often. Windows has the feature to keep files created with the "TEMPORARY" attribute in RAM rather flushing them to disk.

I am concerned of scenarios, where users edit huge code files of several MBs, which are written to disk second after second, which in the end might help wearing out SSDs. Maybe a bit too paranoid, but I think it's worth thinking about, anyway. I would like git diff to support stdin to provide the view's content to avoid tempfiles at all.

rchl: Maybe most benefits are lost when using it like that, but there might be still some. At least it shouldn't be worse than creation file manually. It at least picks the appropriate location for the file automatically.

deathaxe:
I just found `git diff` returning empty string on `stdout` with exit code 128, if the view's file is not part of the git-index. Therefore I extended the `git_handler.run_command()` code to forward the exit code from git to all resolvers, but now I have two choices, which may bring us back to the git_gutter_show_diff discussion.

1. I can modify `decode_diff(...)` to simulate a diff output of "@@ -0,0 +1,_lines_ @@" to keep compatibility with current `git_gutter_show_diff` command.
2. I can set an attribute in `decode_diff(...)` to indicade the "file not in git-index" state and change the GitGutterShowDiff class to handle it correctly.

IMO the second choice has two advantages.
1. Number of lines doesn't need to be counted to generate the diff string as in choice 1.
2. The heurictic method `_are_all_lines_added(...)`, which counts them again, can be removed, too.

The contra is this makes it impossible to create a second pull request for the reverted `git_gutter_show_diff` as it would depend on this one.

_Btw.:_ I checked out code of V1.3.0 and found that markers on untracked or ignored files are shown only if the `git_gutter_show_diff` is set true. Otherwise no markers are added at all, just as I expected. So my recent suggestion would only revert behaviour to the state of V1.3.0

deathaxe:
I finally reverted the branch to nearly approved state, because I found out today, that git has problems with line ending conversion in some situations. This means using git directly as probosed, will not work for all users and results in the whole file being marked modified, even if it is not. 

That's why I reverted those changes and would like to close this request, if no more issues or changes are desired by one of you.

git_gutter_handler.py
```diff
@@ -285,33 +292,44 @@ def process_diff_line_change(self, line_nr, diff_str):
...
+                git returns empty output to stdout.
+                """
+                return bool(results[1])
 
             args = [
                 settings.git_binary_path,
                 '--git-dir=' + self.git_dir,
                 '--work-tree=' + self.git_tree,
-                'ls-files', '--other', '--exclude-standard',
+                'ls-files', '--other', '--exclude-standard'
```
rchl: nit: It's a good practice to keep trailing commas from the perspective of git history. Adding more arguments won't require changing this line then.

git_gutter_handler.py
```diff
@@ -285,33 +292,44 @@ def process_diff_line_change(self, line_nr, diff_str):
...
-                    decoded_results = results.decode(encoding.replace(' ', ''))
-                except UnicodeError:
-                    decoded_results = results.decode("utf-8")
-                return (decoded_results != "")
+                """Determine if view's file is in git's index.
+
+                If the view's file is not part of the index
+                git returns empty output to stdout.
+                """
+                return bool(results[1])
```
rchl: Why not results[0]?

deathaxe: Regression! I returned git exit code in the results in the reverted changes, when running git diff directly against the file. This is obiously wrong and you are right as usual.

git_gutter_handler.py
deathaxe: Ok, not a big deal. I removed it only, because in other argument lists last lines have no comma, too.

PR #338, opened 2016-11-19
Let user show/hide markers on new, ignored and untracked files

deathaxe:
The documentation of `show_markers_on_untracked_file` setting quite clearly says: "GitGutter shows icons for new files and ignored files." 

GitGutter 1.3.0 did behave exactly like documented.
> git diff is called only for files, which are not marked untracked or ignored.

GitGutter 1.4.0 shows '+' markers on each line for any file within a git repository even with `show_markers_on_untracked_file: false`.
> git diff is called first and then a check is performed, whether the file might be untracked or ignored.

This pull request intends to revert the behaviour to as it was on 1.3.0 to satisfy documentation.

`show_markers_on_untracked_file: false`
	-> do not show any marker on ignored, untracked or new files.

`show_markers_on_untracked_file: true` (Default)
	-> show '*' for untracked files
	-> show 'x' for ignored files
	-> show '+' for new files

### DIFF
--- a/GitGutter.sublime-settings
+++ b/GitGutter.sublime-settings
@@ -57,7 +57,7 @@
   "ignore_whitespace": "none",
 
   // Add a special marker on untracked files
-  "show_markers_on_untracked_file": false,
+  "show_markers_on_untracked_file": true,
 
   // Add --patience switch to git diff command. See
   // http://bramcohen.livejournal.com/73318.html for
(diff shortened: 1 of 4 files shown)


r-stein:
We should also disable to popup in such files

deathaxe:
I rebased and squashed the branch to the latest master.

PR #339, opened 2016-11-20
Implements coala-ci

mamat-rahmat:
Adds linters for file types : .py, .yml, .rb, .css, .json, .md
More information about coala : https://coala.io/


git_gutter_popup.py
```diff
@@ -203,15 +203,14 @@ def is_button_enabled(k):
...
-    for css_file in css_files:
-        try:
-            css_contents.append(sublime.load_resource(css_file))
-        except OSError:
-            print("Error while loading '{0}'.".format(css_file))
-    css = "\n".join(css_contents)
+    # load the user css file
+    css = sublime.load_resource("Packages/GitGutter/gitgutter_popup.css")
+    try:
+        user_css = sublime.load_resource("Packages/User/gitgutter_popup.css")
```
jayvdb: I dont have an opinion on switching from `find_resources` to `load_resource`, however IMO this should be

```js
    try:
        css += "\n{0}".format(
            sublime.load_resource("Packages/User/gitgutter_popup.css"))
    except ...:
        ...
```

Less string concatenations.

git_gutter.py
```diff
@@ -19,6 +19,7 @@
 
 
 class GitGutterCommand(sublime_plugin.TextCommand):
+
```
jayvdb: `pycodestyle` has an option to disable the rule which requires a blank line here, if the maintainers do not like having an extra blank line here. (it seems to be a very common style choice; maybe it is intentional disagreement with PEP8)

README.md
```diff
@@ -10,26 +11,30 @@ A sublime text 2/3 plugin to show an icon in the gutter area indicating whether
 
 You can install via [Sublime Package Control](http://wbond.net/sublime_packages/package_control):
 
-  * [GitGutter](https://sublime.wbond.net/packages/GitGutter) - Release based
```
jayvdb: is `*` or `-` the preferred style of the developers here?  the linter can be configured to support either.

git_gutter_popup.py
deathaxe: Actually this pull request breaks at least the already applied pull request #336 and maybe more.
Other still open pull requests would break, too.

This brings neither functionality nor does it fix any bugs, so wait until all other pr's will be merged.

git_gutter.py
deathaxe: pydocstyle wants a docstring here by default.

deathaxe:
If someone is willing to perform such a mass of changes only to satisfy linting, lets do this, if ongoing developments have finished. Maybe just before a next release. Otherwise this makes current  developements very hard.

git_gutter.py
jayvdb: A blank line is still required after the docstring, if one is present.

git_gutter_popup.py
jayvdb: ya, ok, so this part of the change needs to be reverted.  I wasnt sure, but pull request #336 shows why code shouldnt be changed in order to solve linting rules.

mamat-rahmat:
@deathaxe Fully understable. Lets wait until its ready.

deathaxe:
SublimeLinter with pep8, pyflakes and pydocstyle will be applied to future code changes.

Hard line breaks in Markdown won't be applied as they break line wrapping on the github.

jayvdb:
> Hard line breaks in Markdown won't be applied as they break line wrapping on the github.

Could you clarify what you mean?

Here is how the hard line breaks render in GitHub:

https://github.com/mamat-rahmat/GitGutter/blob/8e218fbdc368588b9c97a80fd244be45e9c34402/README.md

I dont see any breakages.

rchl:
> SublimeLinter with pep8, pyflakes and pydocstyle will be applied to future code changes.

How will you ensure that? Pull requests come from external contributors and you can't know if they ensured that the code meets all the style checks.

It's good to have automatic checking of commits on the github level already.

Locally nobody prevents any linter interface you want for style checking.

rchl:
@mamat-rahmat This plugin does not include any ruby or yaml files. Please don't linter configuration  for those. Keep it as simple as possible.

jayvdb:
> @mamat-rahmat This plugin does not include any ruby or yaml files.

There is `release_messages/src/build.rb` , and this patch introduces `.travis.yml`.

deathaxe:
Did my best to follow the linter rules as close as possible where useful. Some rules applied or enforced by linters are just crap. As there are only few contributions by only few people I don't see any sense in adding support for coala-ci or something like that. I don't like adding stupid comments to disable linter warnings. Furthermore this PR is heavily outdated and therefore closed.

Issue #341, opened 2016-11-22
Release Schedule


jisaacks:
@r-stein @rchl 

Hey guys, I have been really appreciating the all the hard work you have been doing. In light of all the recent activity. I wanted to discuss on a plan for managing everything.

What do you think of always having an open `issue` for the next release, where we can keep track of what is going to be included in the next release, the changelog etc. It would be a central communication hub for everyone to communicate on progress for next releease. Then once everyone approves the release. I can cut it. I would then be fine with y'all merging PRs as you see fit, as long as we are adding to the changelog and all in communication.

This plan may suck, so let me know what y'all think or any modifications to this plan, of if you just think it's bad idea, etc. Let me know! Thanks.

r-stein:
Using milestones seems to be a good idea to group PRs and issues for releases. Additionally creating an issue per release isn't bad either.

rchl:
I'm all up for trying it.

jisaacks:
Do y'all have merge rights?

Let's try this, if one of you approves, the other feel free to merge (prefer rebase.) That way you are always both signing off. Let's try that and see how that works.

And lets try to add milestones, and an issue tracking each release if we can. I'll still cut the releases.

rchl:
I think I do. At least I see the merge button.

As for "prefer rebase". I hate including fixup! commits or commits that just fix issues in the pull request in the final history. So would prefer squashing all commits (or at least do it when it makes sense).

#331 is one example where I would prefer squashing as it has some commits that fix earlier problem of the pull request or just adds changes that are later removed.

But I guess common sense should be applied on a case basis...

jisaacks:
> So would prefer squashing all commits (or at least do it when it makes sense).

My only reasoning against squashing, is I like to give contributors, higher numbers in the commit graph if they did a lot of work. If someone wrote 50 commits, and we squash it to 1 commit, it puts them low on the commit graph, when in fact they helped out more than what the graph portrays.

> #331 is one example where I would prefer squashing as it has some commits that fix earlier problem of the pull request or just adds changes that are later removed.

If it makes sense to squash it or merge, I am fine. I just prefer rebasing if possible because it usually makes the tree a little cleaner. But isn't always worth it.

Just use your own judgement/discretion.

> But I guess common sense should be applied on a case basis...

Exactly! :)

jisaacks:
[...]

PR #344, opened 2016-11-27
Securing git process call to avoid git zombies.

deathaxe:
Reading the documentation of python's `subprocess` module, I found some recommendations, which I packed into this pull request. What do you think about it?

### Python says:

So maybe we should use it.

Added a timeout of 30s.

So print the exception text to console and handle as empty output.

### DIFF
--- a/git_gutter_handler.py
+++ b/git_gutter_handler.py
@@ -1,10 +1,11 @@
 import os
-import subprocess
 import re
 import codecs
 import tempfile
 import time
 from functools import partial
+from subprocess import (
+    Popen, TimeoutExpired, PIPE, STARTUPINFO, STARTF_USESHOWWINDOW)
 
 import sublime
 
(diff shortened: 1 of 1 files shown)


rchl:
Sorry, but you are not following the process. At least one person should approve the changes before they are merged - https://github.com/jisaacks/GitGutter/issues/341#issuecomment-262311323.

Please don't just merge fixes without review.

git_gutter_handler.py
```diff
@@ -399,17 +400,24 @@ def run_command(args):
...
+            try:
+                stdout = b''
+                startupinfo = None
+                if os.name == 'nt':
+                    startupinfo = STARTUPINFO()
+                    startupinfo.dwFlags |= STARTF_USESHOWWINDOW
+                proc = Popen(
+                    args=args, stdin=PIPE, stdout=PIPE, stderr=PIPE,
+                    startupinfo=startupinfo)
+                stdout, stderr = proc.communicate(timeout=30)
```
rchl: timeout seems python3 only. Does this break Sublime2 support?

Issue #256, opened 2015-10-12
Custom icons


monkeez:
Hey there, love the plugin. I was just wondering if it would be possible to support custom icons for GitGutter? Or is this something that's not possible with Sublime plugins?

Thanks

PR #347, opened 2016-12-01
Add high-res icons

### LINKED ISSUE #199: Icons are very small on high-res display
This computer has a stupidly large screen resolution, which means that the icons are too small to be noticeable really:

image

nwoltman:
This PR adds a set of high-res ("2x") icons and a `high_res_icons` setting that users can set to `true` to tell GitGutter to render the high-res icons in the gutter instead of the default icons.

Fixes #199


jisaacks:
Is this something that only effects window/linux? I have never had an issue on OSX with retina/HiDPI displays.

nwoltman:
Yes it's only an issue on Windows/Linux.

deathaxe:
Basically an setting about high-res may not be a bad idea, but I ask myself, whether this could be handled by support of custom icon packages like suggested in issue #256 in a more general way.

rchl:
This is a workaround for a bug in Sublime. I would rather see you putting effort in convincing Sublime authors to fix it. :) It's also not ideal that a workaround requires a new setting that user has to find and toggle. If we would really go with a workaround, I'd try to find a way to autodect hiDPI mode and platform and apply such workaround automatically.

And assuming that this is fixed in Sublime later, such workaround would then backfire by making all icons too big.

nwoltman:
@rchl It has been over 3 years since Sublime introduced support for hiDPI screens. At this point it seems extremely unlikely that it's going to get fixed anytime soon.

Actually, that makes me think that perhaps there is no "bug" at all in Sublime. It may simply be a difference in screen resolution. I have a 4K screen (3840x2160) so the GitGutter icons look tiny on my screen. Someone like @jisaacks who has a Mac Retina screen might have a resolution of 2880x1800, so the icons could look fine at that resolution.

If that is the case (which is starting to seem pretty likely seeing as how #199 was opened by someone with a 3200x1800 resolution screen), autodetecting hiDPI is not a solution since there are a wide variety of resolutions that are considered hiDPI. Therefore, it would be best to let users opt in to the "high_res_icons" setting based on whether or not the icons are too small on their particular screen.

rchl:
IMO it's not about resolution but about pixel density.
Whether 1920x1200 or 3840x2160, if the pixel density would be the same in both cases, the marker would be equally well visible. All elements of the UI would have the same relative scale in both of those cases.

With higher pixel density (on retina or you 4k screen), where there a lot more more pixels packed per inch, to keep things readable, app needs to scale resources according to pixel density (or device pixel ratio). That is what Sublime is not doing here due to a bug. It  does it correctly on Mac and it does it correctly for it's "native" UI elements. It just forgot to apply same logic to images specified through add_regions API.

rchl:
See in this image: 
scaling

Those native UI/Theme graphic elements marked with green outline are handled properly. These are scaled according to the device's pixel ratio. They become ugly and blurry in effect but they have proper size at least.

The git gutter marker is not scaled so it's perfectly sharp, but too small.

The fix might be simple 5 minutes job. That fact that it didn't happen yet might be because it didn't get enough attention.
And to be fair, Sublime development isn't staggeringly fast.

nwoltman:
Yes you're right, I forgot it's the user's pixel density that's important because the icons should be scaled by the same ratio as the rest of the UI.

Also, I just verified that Sublime does scale gutter icons on Mac but not on Windows/Linux.

Anyway, we have no idea when or if this will be fixed in Sublime, so until that point, there should be something in GitGutter to provide Windows/Linux with a better experience.

rchl:
Superseded by @358 so closing.

deathaxe:
Since PR #361 added support for custom icon packages with a `Default HIDPI.gutter-theme` to support 4k/8k displays this PR might not be required anymore and therefore closed.

nwoltman:
@deathaxe The custom icon support looks great! I wanted to try it out but it looks like the [GitGutter-Edge](https://sublime.wbond.net/packages/GitGutter-Edge) package doesn't exist anymore :(

deathaxe:
We want to change the strategy to use the pre-release feature of package control and therefore remove GitGutter-Edge as the `-Edge` has some pitfalls. The different package name makes some *.sublime-menu and *.sublime-commands not working.

The idea exists for a while now but in the last few days things moved on quite fast. 

You can 
1. pull directly from master branch for now or
2. install normal GitGutter package and give us some days to make up our mind for a final tagging scheme for pre releases.

Add the following setting to your `Package Control.sublime-settings` to tell it to install pre releases

```js
	"install_prereleases":
	[
		"GitGutter"
	],
```

nwoltman:
Looks great on master! Thanks @deathaxe 👍

PR #348, opened 2016-12-05
Securing git execution.

deathaxe:
The documentation about python's `subprocess` module makes some recommendations
about how to handle subprocess calls, which are applied by this commit.

### Python says:

avoid deadlocks due to any of the other OS pipe buffers filling up and blocking
the child process.

So maybe we should use it.

properly a well-behaved application should kill the child process and finish
communication.

Would be nice but is not supported by python 2.6 and therefore does not work
with SublimeText 2.

trying to execute a non-existent file. Applications should prepare for OSError
exceptions.

Catch any exception, print its text to console and handle as empty output. Finally
resolve the Promise to avoid memory leaks due to unresolved zombie promises.

### DIFF
--- a/git_gutter.py
+++ b/git_gutter.py
@@ -1,6 +1,7 @@
 import sublime_plugin
 
 try:
+    from .git_gutter_settings import settings
     from .git_gutter_handler import GitGutterHandler
     from .git_gutter_compare import (
         GitGutterCompareCommit, GitGutterCompareBranch, GitGutterCompareTag,
@@ -9,6 +10,7 @@ try:
     from .git_gutter_popup import show_diff_popup
     from .git_gutter_show_diff import GitGutterShowDiff
 except (ImportError, ValueError):
+    from git_gutter_settings import settings
     from git_gutter_handler import GitGutterHandler
     from git_gutter_compare import (
         GitGutterCompareCommit, GitGutterCompareBranch, GitGutterCompareTag,
@@ -26,7 +28,7 @@ class GitGutterCommand(sublime_plugin.TextCommand):
         self.show_diff_handler = None
 
     def is_enabled(self, **kwargs):
-        return self.is_valid_view
+        return self.is_valid_view and bool(settings.git_binary_path)
 
     def run(self, edit, **kwargs):
         if not self.git_handler:
--- a/git_gutter_handler.py
+++ b/git_gutter_handler.py
@@ -1,11 +1,10 @@
 import os
+import subprocess
 import re
 import codecs
 import tempfile
 import time
 from functools import partial
-from subprocess import (
-    Popen, PIPE, STARTUPINFO, STARTF_USESHOWWINDOW)
 
 import sublime
 
@@ -402,16 +401,16 @@ class GitGutterHandler(object):
             """Start git process and forward its output to the Resolver."""
             try:
                 if os.name == 'nt':
-                    startupinfo = STARTUPINFO()
-                    startupinfo.dwFlags |= STARTF_USESHOWWINDOW
+                    startupinfo = subprocess.STARTUPINFO()
+                    startupinfo.dwFlags |= subprocess.STARTF_USESHOWWINDOW
                 else:
                     startupinfo = None
-                proc = Popen(
-                    args=args, stdin=PIPE, stdout=PIPE, stderr=PIPE,
-                    startupinfo=startupinfo)
+                proc = subprocess.Popen(
+                    args=args, stdin=subprocess.PIPE, stdout=subprocess.PIPE,
+                    stderr=subprocess.PIPE, startupinfo=startupinfo)
                 stdout, stderr = proc.communicate()
             except Exception as exception:
-                print("GitGutter failed to run git: " + str(exception))
+                print('GitGutter failed to run git: %s' % exception)
                 stdout = b''
             finally:
                 resolve(stdout)
--- a/git_gutter_handler.py
+++ b/git_gutter_handler.py
@@ -1,10 +1,11 @@
 import os
-import subprocess
 import re
 import codecs
 import tempfile
 import time
 from functools import partial
+from subprocess import (
+    Popen, PIPE, STARTUPINFO, STARTF_USESHOWWINDOW)
 
 import sublime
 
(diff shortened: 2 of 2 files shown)


git_gutter_handler.py
```diff
@@ -1,10 +1,11 @@
 import os
-import subprocess
 import re
 import codecs
 import tempfile
 import time
 from functools import partial
+from subprocess import (
+    Popen, PIPE, STARTUPINFO, STARTF_USESHOWWINDOW)
```
rchl: Importing STARTUPINFO and probably one after will throw on non-Windows.

(IMO better to just import whole subprocess like before.)

git_gutter_handler.py
```diff
@@ -399,17 +400,21 @@ def run_command(args):
...
+                    startupinfo = STARTUPINFO()
+                    startupinfo.dwFlags |= STARTF_USESHOWWINDOW
+                else:
+                    startupinfo = None
+                proc = Popen(
+                    args=args, stdin=PIPE, stdout=PIPE, stderr=PIPE,
+                    startupinfo=startupinfo)
+                stdout, stderr = proc.communicate()
+            except Exception as exception:
+                print("GitGutter failed to run git: " + str(exception))
```
rchl: I think it's generally preferred to use string formatting in python:
```
print("... %s" % exception).
```

git_gutter_handler.py
```diff
@@ -399,17 +400,21 @@ def run_command(args):
...
+                if os.name == 'nt':
+                    startupinfo = STARTUPINFO()
+                    startupinfo.dwFlags |= STARTF_USESHOWWINDOW
+                else:
+                    startupinfo = None
+                proc = Popen(
+                    args=args, stdin=PIPE, stdout=PIPE, stderr=PIPE,
+                    startupinfo=startupinfo)
+                stdout, stderr = proc.communicate()
+            except Exception as exception:
```
rchl: To be fair why catch any exception at all if you are not doing any special handling but just re-printing it?

git_gutter_handler.py
jisaacks: Hey Python, whatever happen to 

> There should be one-- and preferably only one --obvious way to do it.

😂

deathaxe:
A TODO I learned from this for future is to handle git errors itself.

git_gutter_handler.py
deathaxe: And which of all that solutions is agreed by everybody? I ask kindly to finish the discussion on such details quickly as the current small PR's were meant to be the base for two bigger ones, which intend to fix some real open issues and implement some requested features.

The % is used in other modules, too. But your last suggestion does not look so bad, too ;-) as I also  try to avoid format strings where possible in other projects.

rchl:
> 2. handle timeouts
> Would be nice but is not supported by python 2.6

You could still handle it in ST3 bu using conditional block based on a constant defined earlier. Like in git_gutter_events.py, for example:
```
ST3 = int(sublime.version()) >= 3000

...

if ST3:
  process.communicate(timeout=15)
else
  process.communicate()
```

git_gutter_handler.py
```diff
@@ -12,11 +12,20 @@
...
+    from subprocess import TimeoutExpired
+
+    _HAVE_TIMEOUT = True
+
 except (ImportError, ValueError):
     import git_helper
     from git_gutter_settings import settings
     from promise import Promise
 
+    class TimeoutExpired():
```
rchl: Maybe makes sense to inherit from "Exception". Even if it won't really be triggered.

git_gutter.py
```diff
@@ -26,7 +28,7 @@ def __init__(self, *args, **kwargs):
         self.show_diff_handler = None
 
     def is_enabled(self, **kwargs):
-        return self.is_valid_view
+        return self.is_valid_view and bool(settings.git_binary_path)
```
rchl: That might be pre-existing flaw, but if you are just accessing a 'git_binary_path' property of settings like this, you don't have guarantee that it was initialized (that load_settings()) was called. Maybe you should add `__get__` descriptor on the Settings class that will ensure that settings are loaded before returning the value.

git_gutter_handler.py
deathaxe: I expected it to catch anything then, just as I would write `except Exception`.

git_gutter_handler.py
```diff
@@ -12,11 +12,20 @@
     from . import git_helper
     from .git_gutter_settings import settings
     from .promise import Promise
+    from subprocess import TimeoutExpired
```
deathaxe: This try ... except block is to check, if python 3 is used with fallback to 2 only, isn't it? If any of those imports fails it's most likely due to python 2. If it was due to missing object, it would fail in except block, too, which would cause the whole module not to work, right?

git_gutter_handler.py
```diff
@@ -12,11 +12,22 @@
     from . import git_helper
     from .git_gutter_settings import settings
     from .promise import Promise
+
```
rchl: unnecessary addition

git_gutter.py
rchl: It's just not correct. It probably only works by chance because GitGutterEvents trigger settings load when using settings.get('foo'). But it's not a new problem so can keep it for now.

deathaxe: Just to ensure I don't miss something:

`settings.load_settings()` is called by `git_gutter_settings.plugin_loaded()` which itself is called by ST3 as soon as API is ready and on loading the module by ST2. No API call like `sublime.load_settings()` must be called before this point anyway. So even a getter which tries to load settings would most likely fail at this point anyway. This change activly disables git_gutter command if API is not yet ready and therefore no settings are available anyway and thus prevents running git_handler with an empty args[0], which would cause an exception. SublimeText can fire events, but GitGutter ignores them until ready. And if no git was found at all, GitGutter shows its message box triggered in settings.load_settings() and is disabled then. 

So seriously: There are very obvious issues with settings and how it stores values and reloads them from the backend files, but I can't see this to be one of them. 

In the future I could imagine to try to update `git_binary_path` if git command fails with error 2 to check activly for changed location, but this is not what this PR intends to add.

This PR intents to prevent exceptions from resolving the Promise, which is created to run the git process. Issues with the settings module should go to a seperate one, I think.

PR #349, opened 2016-12-05
Fix issue: GitGutter ignores on_modified() event in ST2, because view visibility check fails.

deathaxe:
We need to use the view id to check the view is the active one in a group.

### DIFF
--- a/git_gutter_events.py
+++ b/git_gutter_events.py
@@ -118,6 +118,10 @@ class GitGutterEvents(sublime_plugin.EventListener):
 
         Only an active view of a group is visible.
         """
-        w = view.window()
-        return any(view == w.active_view_in_group(g)
-                   for g in range(w.num_groups())) if w is not None else False
+        window = view.window()
+        if window:
+            view_id = view.id()
+            for group in range(window.num_groups()):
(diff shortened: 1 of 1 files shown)


rchl:
What does missing on_modified() event has to do with this change?
And why do we have to use ID for comparison rather than comparing view object itself. It's not explained by the commit message.

deathaxe:
I tried GitGutter with Sublime Text 2 yesterday and recognized gutter icons not being updated while editing. In the end I saw the `is_view_visible()` method returning always `False`. Printing the view object showed up, they have different addresses. I changed the code, like proposed and it worked for ST2 and 3 as well.

r-stein:
I also hat a similar issue in ST2. In addition you should not assume that there is an active view in the group, such that the id() function call raises an exception.

rchl:
>> What does missing on_modified() event has to do with this change?

> I tried GitGutter with Sublime Text 2 yesterday and recognized gutter icons not being updated while editing. 

Sure, but you agree that this has nothing to do with ST2 not handling on_modified()? :) Just asking for correcting the commit message. :)

deathaxe:
Is it ok to merge?

Issue #264, opened 2016-01-13
Broken Menu File


wbond:
https://github.com/jisaacks/GitGutter/blob/master/Main.sublime-menu#L3-L56

These menu entries add blank top-level menus on OS X with ST3 build 3095.

jisaacks:
Thanks for pointing this out, Will. I am currently on 3083 it appears to be working in this build. I Wonder what has changed. I will update ST and take a look when I can.

rchl:
@wbond Do you still see this problem? I don't see anything like that on macOS Sierra. The file you pointed to doesn't seem to have any problems to me. But then it always points at master so it might not longer show what you tried to show.

PR #350, opened 2016-12-05
Fix Issue #264 "Broken Menu File"

deathaxe:
Even though all menu items appear to be ok it might be a clean solution to
add the caption to all existing items, too. This at least avoids some issues
and is well practice in most plugins anyway.

### DIFF
--- a/Main.sublime-menu
+++ b/Main.sublime-menu
@@ -1,5 +1,7 @@
 [
     {
+        "caption": "View",
+        "mnemonic": "V",
         "id": "view",
         "children":
         [
@@ -42,17 +44,20 @@
         ]
     },
     {
+        "caption": "Goto",
+        "mnemonic": "G",
         "id": "goto",
         "children":
         [
             {
-                "caption": "Previous Change",
+                "caption": "Goto Previous Git Difference",
                 "command": "git_gutter_prev_change"
             },
             {
-                "caption": "Next Change",
+                "caption": "Goto Next Git Difference",
                 "command": "git_gutter_next_change"
-            }
+            },
+            { "caption": "-" }
         ]
     },
     {


Main.sublime-menu
```diff
@@ -1,5 +1,7 @@
 [
     {
+        "caption": "View",
```
rchl: I'm not sure about that change. If we can't reproduce any problem maybe it was fixed in Sublime builds?

Specifying caption seems less generic and more error prone because it will override existing caption. And if, for example, Sublime Text would decide to slightly rename its menu item caption, the plugin would keep overriding it here.

Main.sublime-menu
```diff
@@ -42,17 +44,20 @@
...
         "children":
         [
             {
-                "caption": "Previous Change",
+                "caption": "Goto Previous Change",
                 "command": "git_gutter_prev_change"
             },
             {
-                "caption": "Next Change",
+                "caption": "Goto Next Change",
```
rchl: Isn't "Goto" redundant given that it's in the "Goto" menu group?

I guess I can't make a point that either one would be more consistent, but I don't think I see advantage of this change.

jisaacks: I agree. It's redundant.

Main.sublime-menu
jisaacks: I don't really understand how this will fix anything? If the `id` did not work why would the caption?

Did you confirm that it was broken and this fixes it?

Main.sublime-menu
r-stein: The other entries also start with "Goto", so this would be reasonable.
We may also add a link to GitGutter, e.g. "GitGutter: Goto Next Change", because it is not clear that the change is related to the git state. In addition we may disable that command outside of git files and if no change is available.

Main.sublime-menu
deathaxe: After wbond's critics I had a look into some packages providing Main.sublime-menu. All of them provide a caption as well so I found it "state of the art" to do so, too, even though the empty menu item might have been an issue in Sublime Text itself. Does it hurt to add it?

Main.sublime-menu
deathaxe: r-stein gives the answer and the reason for my change. I never realized this menu entry and had a aha effect, when I saw in in the Main.sublime-menu. The first impression on this menu was just ... those items are missing something. @r-stein your might be right, but must we add GitGutter or is there a more elegant alternitive?

r-stein: We could also rephrase it and include the information, that is targets the git stats, e.g. "Goto Next Git Change" or "Goto Next Git Difference".

deathaxe: KDiff3 calls it `Goto Next Delta`, which might used as `Goto Next Git Delta`, but `Goto Next Git Difference` sounds good, too. Any further suggestions?

Main.sublime-menu
rchl: I won't insist on either having or not having Caption. I just think that it's not ideal to override caption of existing menu items, which seems what Sublime is doing. If it would only use provided caption if menu wouldn't exist already then it would maybe make more sense.

As it is now if, let's say, someone would translate the whole menu locally, then he wouldn't it to be overridden my some plugin.
But maybe I'm worrying too much about corner cases.

Main.sublime-menu
r-stein: While Delta is common as difference operator in mathematics and also often used in computer science, I would not assume everybody understands it. 
I vote for difference.

jisaacks: We could just use "Goto Next Δ" haha. But in all seriousness, git itself calls them _diffs_ so it might make more since to use the same vocabulary. However, I personally like "Goto Next Change" myself. But I am OK with whatever most people agree on.

deathaxe: The point is to show the user what changes are meant. I would never have expected those menu items to represent a function of GitGutter.

Issue #189, opened 2014-09-10
binary files


krya:
if you try to open a binary file gitgutter raises an exception

``` python
Traceback (most recent call last):
  File "./sublime_plugin.py", line 339, in run_
  File "./git_gutter.py", line 45, in run
  File "./view_collection.py", line 51, in diff
  File "./git_gutter_handler.py", line 156, in diff
LookupError: unknown encoding: Hexadecimal
```

sublime 2 on linux here.

KevinLi:
ST3 still returns "Hexadecimal" in the hexadecimal view.
I've just made a pull request that should fix this.

rchl:
@KevinLi Can't see your pull request :)

KevinLi:
Me neither!

KevinLi:
Me neither! I have no idea what's going on!
It's been nearly two years; does the bug still exist?

rchl:
I think it still exists looking at the code. There is no rewriting of encoding from 'Hexadecimal' to 'hex'.

jisaacks:
> Me neither! I have no idea what's going on!

Probably, you deleted the repo from your own account, which the fork was based on. So the fork went away?

PR #351, opened 2016-12-05
Fix Issue #189: Disable GitGutter for binary files and scratch views.

deathaxe:
There are several situations running GitGutter might be useless or disturbing, so it should keep disabled.

### 1 binary files
As SublimeText's hexadecimal encoded representation of a binary file wouldn't
be saved to disk in a form git diff could handle it and there is no reason to show
markers for those files, the whole git_gutter command is disabled if a view
returns 'Hexadecimal' encoding.

### 2 scratch / readonly / widget views
Views marked as scratch or readonly are most likely not intended to represent normal files
so handle them like widgets.

### 3 unattached views
SublimeLinter disables itself for views with no valid window() attached and states it as possible way to detect widgets for sure in some builts of SublimeText.

### 4 SublimeREPL
The sublimeREPL plugin sets a view setting `repl` to mark a view as output. SublimeLinter uses this setting to disable itself, so GitGutter should do so, too.

### Where to check the view?
As most activity of GitGutter is triggered by events, it might be efficient to use the EventListener to check if a view is valid and disable event handling at all if needed. Checking the view after each key press by `on_modified` may be not required, so check only on activition and after saving a file.

### DIFF
--- a/git_gutter.py
+++ b/git_gutter.py
@@ -1,4 +1,4 @@
-import sublime_plugin
+from sublime_plugin import TextCommand
 
 try:
     from .git_gutter_settings import settings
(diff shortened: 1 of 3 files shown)


git_gutter.py
```diff
@@ -20,31 +20,53 @@
...
+        if not bool(settings.git_binary_path):
+            valid = False
+        # Don't handle unattached views
+        elif view.window() is None:
+            valid = False
+        # Don't handle scratch or readonly views
+        elif view.is_scratch() or view.is_read_only():
+            valid = False
+        # Don't handle widgets
+        elif view.settings().get('is_widget') is True:
```
rchl: Style nit: I think `if not Foo` reads better (more natural) than `Foo is False` / `Foo is None`.
And `if Foo is True` can just be `if Foo`.

git_gutter.py
```diff
@@ -20,31 +20,53 @@
...
+        view.settings().set('git_gutter_enabled', valid)
+        return valid
 
     def run(self, edit, **kwargs):
-        if not self.git_handler:
-            self.git_handler = GitGutterHandler(self.view)
+        """API entry point to run the `git_gutter` command."""
         if not self.show_diff_handler:
             self.show_diff_handler = GitGutterShowDiff(
                 self.view, self.git_handler)
```
rchl: I think it makes sense to also create this in init. It doesn't do any expensive stuff on init. It was here only because it depended on GitGutterHandler.

git_gutter.py
```diff
@@ -77,45 +99,54 @@ def _handle_subcommand(self, **kwargs):
...
+class GitGutterBaseCommand(TextCommand):
+    def is_enabled(self, **kwargs):
+        return self.view.settings().get('git_gutter_enabled', False)
+
+
+class GitGutterShowCompareCommand(GitGutterBaseCommand):
     def run(self, edit):
-        self.view.run_command('git_gutter', {'action': 'show_compare'})
+        self.view.run_command(
+            'git_gutter', {'action': 'show_compare'})
```
rchl: Style nit: That did fit on one line just fine. No need to change it.
I suppose you've done that so that it looks consistent in all classes but that is IMO not necessary. Taking least vertical space is more important.

git_gutter.py
```diff
@@ -77,45 +99,54 @@ def _handle_subcommand(self, **kwargs):
...
 
 
-class GitGutterCompareHeadCommand(sublime_plugin.TextCommand):
+class GitGutterCompareHeadCommand(GitGutterBaseCommand):
     def run(self, edit):
-        self.view.run_command('git_gutter', {'action': 'compare_against_head'})
+        self.view.run_command(
+            'git_gutter', {'action': 'compare_against_head'})
```
rchl: That does fit on one line.

git_gutter.py
```diff
@@ -77,45 +99,54 @@ def _handle_subcommand(self, **kwargs):
...
     def run(self, edit):
         self.view.run_command(
             'git_gutter', {'action': 'compare_against_branch'})
 
 
-class GitGutterCompareTagCommand(sublime_plugin.TextCommand):
+class GitGutterCompareTagCommand(GitGutterBaseCommand):
     def run(self, edit):
         self.view.run_command(
             'git_gutter', {'action': 'compare_against_tag'})
```
rchl: Fits one line.

git_gutter.py
```diff
@@ -77,45 +99,54 @@ def _handle_subcommand(self, **kwargs):
...
 
 
-class GitGutterNextChangeCommand(sublime_plugin.TextCommand):
+class GitGutterNextChangeCommand(GitGutterBaseCommand):
     def run(self, edit):
-        self.view.run_command('git_gutter', {'action': 'jump_to_next_change'})
+        self.view.run_command(
+            'git_gutter', {'action': 'jump_to_next_change'})
```
rchl: Fits one line.

git_gutter.py
```diff
@@ -77,45 +99,54 @@ def _handle_subcommand(self, **kwargs):
...
 
 
-class GitGutterPrevChangeCommand(sublime_plugin.TextCommand):
+class GitGutterPrevChangeCommand(GitGutterBaseCommand):
     def run(self, edit):
-        self.view.run_command('git_gutter', {'action': 'jump_to_prev_change'})
+        self.view.run_command(
+            'git_gutter', {'action': 'jump_to_prev_change'})
```
rchl: Fits one line.

git_gutter.py
deathaxe: Yes I did for consistent look to the other ones, but if vertical space is most importand you are right. Will revert those lines.

Issue #281, opened 2016-06-03
Compare against specific commit Enhancement


suiwenfeng:
1. commits listed should come from file change history and order by commit time.
2. filter commit by commit id not only commit message.

jisaacks:
@suiwenfeng makes sense. is it something you'd be interested in working on?

deathaxe:
Should this replace the 'Compare against Commit' or is a new 'Compare against File Commit' command prefered.

As the compare target is saved per directory now, both makes sense, somehow. First solution is easer - needs only two lines of code being changed. Could be included in PR #343

Issue #340, opened 2016-11-21
Noticeable flicker/delay in icons when working with files


sgerace:
So I got the latest update today (1.4.0) and I'm seeing a very noticable flicker/delay from when I update the code and when the gitgutter icons are refreshed. It looks like the state of the lines is not being remembered and instead all of the lines are cleared and refreshed. I can't say for certain that this was caused by the 1.4.0 update, however, I feel like I would have noticed this if it was happening before. The best way to illustrate the issue is with a screen capture:

![flicker](https://cloud.githubusercontent.com/assets/1239058/20485699/c0ead3d2-afca-11e6-9fe5-487dc6864925.gif)

All I'm doing in this video is typing, and I'm not saving the file or refreshing anything. Let me know if there is anything I can do to help diagnose the issue. Thanks!

rchl:
Previously GitGutter was blocking the editor when updating markers so you wouldn't see the markers flickering but you would experience freezes in the editor instead.

I guess the bug is still valid. It can probably be fixed by delaying clearing of markers in some circumstances. But it won't necessarily be an easy fix.

r-stein:
Can you try to remove/outcomment [this line](https://github.com/jisaacks/GitGutter/blob/1.4.0/git_gutter_show_diff.py#L68)?The regions are overwritten by readding them anyway so this should not be necessary.

sgerace:
Sure thing, only I'm not sure where to make this change to have it take effect. I tried looking around for where Package Control installs the source code for the packages I have installed, but wasn't able to find it.

r-stein:
Oh sorry install [PackageResourceViewer](https://packagecontrol.io/packages/PackageResourceViewer), press ctrl+shift+p and run the command "PackageResourceViewer: Open Resource". Aferwards select "GitGutter/git_gutter_show_diff.py" to open the file. Do the change and restart ST to be sure the changes are propagated.

sgerace:
Yep, that seems to have fixed the issue.

r-stein:
ok nice =)

jisaacks:
What if you add a new line, causing the `+` to show up. Then remove the new line. Will the `+` still go away?

sgerace:
Yes, it seems like it is behaving properly:

![flicker-2](https://cloud.githubusercontent.com/assets/1239058/20492337/563a23c2-afe2-11e6-9543-1333ddb4b2bb.gif)

deathaxe:
[...]

PR #354, opened 2016-12-10
Enhancements for the "Compare against ..."

deathaxe:
I'd like to introduce the following changes to GitGutter. I packed several things into this PR as they all belong together, modify the same parts of code and would otherwise make too much work to merge correctly again as they base on each other.

### 1 Infrastructure
The first commits are some kind of infrastructure changes to ease handling of git output and reduce some cross references of git_gutter modules. They mainly intend to reduce required lines of code. They are the base of the following functional changes and additions.

### 2 Show branch name in compared against status message

I was always confuesed about the Comparing against: ... status messages. It only shows HEAD or a certain commit hash instead of the name of a branch selected as compare target. 

#### Issue

Comparing against branch doesn't actually tell GitGutter to compare against the HEAD of the selected branch but only against the currently latest commit. This has two side effects. The branch name is not displayed in the status bar and if the branch forwards, GitGutter won't see that.

#### Solution

Save the whole `refs/heads/...` or `refs/tags/...` strings instead of commit hashes, if _Compare against branch_ or _Compare against tag_ was called.

### 3 Make the status message somewhat smarter

The status message should show for all files within a git repository no matter if the gutter icons show or not. So user can see, if the file is contained in a git repo and which state (ignored/untracked) it has. Furthermore the statistics message itself should always show the 'Regions' or 'Lines' extension after the first visible _inserted/deleted/modified_ item.

### 4 Compare against specific file commit (Issue #281) 

Add a new function to select all commits sorted by author date, which introduced changes to the file displayed in the active view.

### DIFF
--- a/git_gutter_handler.py
+++ b/git_gutter_handler.py
@@ -390,13 +390,40 @@ class GitGutterHandler(object):
         return Promise.resolve(False)
 
     def git_commits(self):
+        r"""Query all commits.
+
+        The git output will have following format splitted by \a:
+            <hash> <title>
+            <name> <email>
+            <date> (<time> ago)
+        """
         args = [
             settings.git_binary_path,
             '--git-dir=' + self.git_dir,
             '--work-tree=' + self.git_tree,
             'log', '--all',
-            '--pretty=%s\a%h %an <%aE>\a%ad (%ar)',
+            '--pretty=%h %s\a%an <%aE>\a%ad (%ar)',
             '--date=local', '--max-count=9000'
+        ]
+        return GitGutterHandler.run_command(args)
+
+    def git_file_commits(self):
+        r"""Query all commits with changes to the attached file.
+
+        The git output will have following format splitted by \a:
+            <timestamp>
+            <hash> <title>
+            <name> <email>
+            <date> (<time> ago)
+        """
+        args = [
+            settings.git_binary_path,
+            '--git-dir=' + self.git_dir,
+            '--work-tree=' + self.git_tree,
+            'log',
+            '--pretty=%at\a%h %s\a%an <%aE>\a%ad (%ar)',
+            '--date=local', '--max-count=9000',
+            '--', self.git_path
         ]
         return GitGutterHandler.run_command(args)
 
--- a/git_gutter_handler.py
+++ b/git_gutter_handler.py
@@ -159,7 +159,8 @@ class GitGutterHandler(object):
 
             try:
                 self.update_git_time()
-                return GitGutterHandler.run_command(args).then(write_file)
+                return GitGutterHandler.run_command(
+                    args=args, decode=False).then(write_file)
             except Exception:
                 pass
         return Promise.resolve()
@@ -226,7 +227,8 @@ class GitGutterHandler(object):
                 self.buf_temp_file,
             ]
             args = list(filter(None, args))  # Remove empty args
-            return GitGutterHandler.run_command(args).then(decode_diff)
+            return GitGutterHandler.run_command(
+                args=args, decode=False).then(decode_diff)
 
         if self.on_disk() and self.git_path:
             return self.update_git_file().then(run_diff)
@@ -399,9 +401,14 @@ class GitGutterHandler(object):
         return GitGutterHandler.run_command(args)
 
     @staticmethod
-    def run_command(args):
-        """Run a git command asynchronously and return a Promise."""
+    def run_command(args, decode=True):
+        """Run a git command asynchronously and return a Promise.
 
+        Arguments:
+            args    - a list of arguments used to create the git subprocess.
+            decode  - if True the git's output is decoded assuming utf-8
+                      which is the default output encoding of git.
+        """
         def read_output(resolve):
             """Start git process and forward its output to the Resolver."""
             try:
@@ -424,7 +431,10 @@ class GitGutterHandler(object):
                 proc.kill()
                 stdout, stderr = proc.communicate()
             finally:
-                resolve(stdout)
+                if decode:
+                    resolve(stdout.decode('utf-8').strip())
+                else:
+                    resolve(stdout)
 
         def run_async(resolve):
             if hasattr(sublime, 'set_timeout_async'):
--- a/git_gutter_handler.py
+++ b/git_gutter_handler.py
@@ -61,6 +61,9 @@ class GitGutterHandler(object):
         os.close(fd)
         return filepath
 
+    def git_time_cleared(self):
+        return self._last_refresh_time_git_file == 0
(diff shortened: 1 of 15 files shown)


git_gutter_handler.py
```diff
@@ -162,7 +162,7 @@ def write_file(contents):
 
             try:
                 self.update_git_time()
-                return GitGutterHandler.run_command(args).then(write_file)
+                return GitGutterHandler.run_command(args, False).then(write_file)
```
rchl: Nit: Would be clearer if you would write `decode=False`.

git_gutter_handler.py
```diff
@@ -229,7 +229,7 @@ def run_diff(_unused):
                 self.buf_temp_file,
             ]
             args = list(filter(None, args))  # Remove empty args
-            return GitGutterHandler.run_command(args).then(decode_diff)
+            return GitGutterHandler.run_command(args, False).then(decode_diff)
```
rchl: Nit: Would be clearer if you would write `decode=False`.

deathaxe:
Unfortunatelly there are some overlaps as this PR is somewhat older. But you are right. Let's check the other one first.

deathaxe:
Rabased the work on new master state and already worked in the handful of tips from @rchl.

r-stein:
The statusbar is a little bit to verbose and a little inconsistent:
```
Comparing against: HEAD, File is modified, Deleted : 1 region, Inserted : 3 lines, Modified: 5
```
We may make the "Comparing against" and "File is modified/added/commited" optional and maybe add the branch and shorted the rest with symbols like:
```
1-, 3+, 5≠
```

deathaxe:
This is exactly the feedback I need! 
1. I aggree with you that all information make it quite long. I already played around with hiding the 'File is modified' and show 'added/staged/commited' states only. This is what it behaves on my local development branch right now.
2. To show 'added/staged' states and efficiently query 'ignored/untracked' files and branch name with a single git call, I created a patch to call 
```python
        args = [
            settings.git_binary_path,
            '--icase-pathspecs',
            'status',
            '--porcelain', '-b', '-s', '-u', '--ignored',
            self._git_path
        ]
```
By now, we need 3 git calls to query branch, ignored state and untracked.
This works pretty well right now, but I am not sattisfied with my handling of the git status yet. This needs some further fine tuning and cleanup. But it is work in progress and is as many things based on my fix for Issue #340  ;-)

For now I may shorten the status message, only. But I need to check, whether ST2 can handle the `≠`

deathaxe:
I remember a dictionary like setting was discussed to enable/disable each element separately was discussed somewhere. May I create such a setting? I did not do so right now, because of compatibility concerns.

r-stein:
My suggestion would be to create a format string, e.g.:

``` js
"status_bar_text": "Branch: {branch}, Comparing against: {compare_target}, File is {file_state} Deleted: {deleted} region, Inserted : {inserted} lines, Modified: {modified}"
```

Why cou can easily change to 

``` js
"status_bar_text": "On {branch}, {deleted}-, {inserted}+, {modified}≠"
```
I am aware that it currently works with several status, but it does not seem to hard to change that.

With a little documentation this seems to be the easiest and most flexible for the users and for us.
We just run

``` py
current_branch = get_current_branch()
status_bar_text.format(branch=current_branch, deleted=deleted, ...)
```

deathaxe:
Sounds great. Will do so.

deathaxe:
Using one format string only will make it impossible to hide certain parts of the status message. The statistics would always show as `0+ 0- 0≠` for a commited file.

So we might split the format string in several parts to avoid such glitches.

How about the following settings and their implementation?

```js
  // Format of the status message.
  // Available variables are:
  //   {repo}     repository name / folder name containing the .git directory
  //   {branch}   checked out branch
  //   {compare}  commit/branch/HEAD the file is compared to
  //   {state}    One of added/modified/staged/untracked/ignored
  "status_bar_text_state": "On {branch}, Comparing against {compare}, File {state}",

  // Modification statistics
  //   Available variables are:
  //   {deleted}  number of deleted regions
  //   {inserted} number of inserted lines
  //   {modified} number of modified lines
  "status_bar_text_stats": "{deleted}- {inserted}+ {modified}≠",
```

r-stein:
I didn't though about that. My idea would be to support the jinja2 template engine instead. We have it as a ST3 dependency anyway for the popup.

Setting:
``` js
"status_bar_text": "On {{branch}}{% if deleted != 0 %}, {{deleted}}-, {% endif %}{% if inserted != 0 %}, {{inserted}}+{% endif %}{% if modified != 0 %}, {{modified}}≠{% endif %}"
```

Code:

``` py
import jinja2
status_bar_text = "On {{branch}}{% if deleted != 0 %}, {{deleted}}-, {% endif %}{% if inserted != 0 %}, {{inserted}}+{% endif %}{% if modified != 0 %}, {{modified}}≠{% endif %}"
tmpl = jinja2.environment.Template(status_bar_text)
tmpl.render(branch="master", deleted=0, inserted=5, modified=12)
```

This gives a less readable setting, but think it is still not to complicated. Especially since the average user knows git and therefore is most likely a programmer himself.

deathaxe:
This is dark magic. Even works for ST2. Just crazy. The pitfall is the very long and maybe complicated template string for some users. Syntax is easy to understand for me, but is it for more genaral users, too? I have no idea how common this template language is.

r-stein:
Me must also adept the [dependencies](https://github.com/jisaacks/GitGutter/blob/master/dependencies.json) to require `python-jinja` for ST2.
Not sure about the average user, but we could add a few example setting strings above.

deathaxe:
Examples are a good idea. Most people will live with default value anyway, I think.

I would change the setting to an array, so the entries can be split into single lines for better readability.

```js
  "status_bar_text": [
    "On {{branch}}",
    "{% if compare != 'HEAD' %}, Comparing against {{compare}}{% endif %}",
    ", File is {{state}}",
    "{% if deleted != 0 %}, {{deleted}}-{% endif %}",
    "{% if inserted != 0 %}, {{inserted}}+{% endif %}",
    "{% if modified != 0 %}, {{modified}}≠{% endif %}"
  ],
```

The array is joint to a single string and passed to jinja2 Template.

deathaxe:
I tend to use `show_status_bar_text:true` instead of `show_status` as it sounds a bit more accurate and may not interfere with the old string setting. What do you mean?

deathaxe:
I rebased this branch on the current master and squashed away the old Enhancement: Status bar ... commit to keep only the commits intended to go to master. Please do a final check, if everything is still fine. If so, I will rebase the master on it.

PR #358, opened 2016-12-18
Add HiDPI icons for Windows/Linux

### LINKED ISSUE #199: Icons are very small on high-res display

nwoltman:
This PR adds a set of high-res ("2x") icons that will be used for Windows/Linux users who have a `dpi_scale` setting of `2` or greater.

Fixes #199 
Supersedes #347


rchl:
As already said, it's a workaround for a bug in Sublime. I have recently reported it at https://github.com/SublimeTextIssues/Core/issues/1522 . I'd love if we wouldn't have to add more complexity and settings to workaround bugs.

But I understand that more pragmatic approach might be actually better for the users so I will not oppose to it. It can always be removed after Sublime bug is fixed.

git_gutter_settings.py
```diff
@@ -20,6 +20,7 @@ def __init__(self):
         self._compare_against_mapping = {}
         # These settings have public getters as they go through more
         # complex initialization than just getting the value from settings.
+        self.dpi_scale = None
```
rchl: You should add this new setting to `GitGutter.sublime-settings` and `README.md`

git_gutter_show_diff.py
```diff
@@ -153,6 +153,10 @@ def _icon_path(self, icon_name):
             if self.view.line_height() > 15:
                 icon_name = icon_name + "_arrow"
 
+        if (sublime.platform() in ['linux', 'windows'] and
+            settings.dpi_scale and settings.dpi_scale >= 2):
```
rchl: It doesn't make sense that the setting takes a number and yet the code handles all values as '2'. It would make more sense to make the setting be a boolean and be named, for example, `large_gutter_icons`. A setting named `dpi_scale` is very ambiguous - not clear what it applies to.

nwoltman: That's what I was going for in #347. Can we just reopen that PR and I'll rename the `high_res_icons` setting to `large_gutter_icons`?

In this PR I was trying to do what you suggested in [this comment](https://github.com/jisaacks/GitGutter/pull/347#issuecomment-264423228) by "autodetecting hiDPI mode" with the `dpi_scale` setting (which is an undocumented setting but some ST users set this in their User Preferences to the DPI scaling factor for their system, such as [this user](https://github.com/jisaacks/GitGutter/issues/199#issuecomment-238432353) from #199).

rchl: By autodetecting I meant doing everything in a way that wouldn't require any settings. So using either Python or system API (through ctypes on Windows I guess) to detect if running in hidpi mode and applying different icons.

I don't know anything about `dpi_scale` but even if it actually does something, I don't think it's a good idea to use some undocumented setting for it.

> Can we just reopen that PR and I'll rename the high_res_icons setting to large_gutter_icons?

Sure, I'll reopen old PR.

deathaxe:
Since PR #361 added support for custom icon packages with a `Default HIDPI.gutter-theme` to support 4k/8k displays this PR might not be required anymore and therefore closed.

Issue #220, opened 2015-04-02
gitgutter-edge empty default preferences file


JeanMertz:
It looks like the default preferences file for `gitgutter-edge` is missing. When opening through the preferences menu, a new (empty) file is created.

Uninstalling and installing non-edge version shows the preference file existing for that version. I presume edge is looking in the wrong directory?

deathaxe:
This is hard to fix as the path in the Main.sublime-menu is fixed. This would require to dynamically create a Main.sublime-menu file for settings.

jisaacks:
I don't want to support gitgutter edge. It doesn't follow any releases. It just points directly to master. It's a way to try out new features. If you want something more solid, install regular gitgutter.

I am also fine with just deprecating gitgutter-edge all together and use pre-releases which sublime text supports.

PR #359, opened 2016-12-26
Fix Issue #220 and add Side-By-Side settings support

deathaxe:
This commit introduces two ApplicationCommands:

1. `git_gutter_edit_settings` is a wrapper for the `edit_settings` command introduced with Sublime Text 3124 to show default and user settings side-by-side.
2. `git_gutter_open_file` is a wrapper for the `open_file` command in Sublime Text used to open settings files.

Both commands expand the variable ${package} to whatever the real local package name of GitGutter is and makes those commands work even if the package was renamed. By default it expands to 'GitGutter' but works with 'GitGutter-Edge', too. This fixes issue #220, even though @jisaacks doesn't want to support Edge anymore.

The second task of those wrapper commands is to make use of `edit_settings` without breaking backward compatibility for older Sublime Text builts.

The three required `edit_settings` commands are added to the command pallet and main menu. The `open_file` commands make use of ${platform} to make the menu file smaller.

### DIFF
--- a/Default.sublime-commands
+++ b/Default.sublime-commands
@@ -30,5 +30,59 @@
     {
         "caption": "GitGutter: Compare Against Tag",
         "command": "git_gutter_compare_tag"
+    },
+    {
+        "caption": "Preferences: GitGutter Settings",
+        "command": "git_gutter_edit_settings",
+        "args": {
+            "base_file": "${packages}/GitGutter/GitGutter.sublime-settings",
+            "default": "// GitGutter Settings - User\n{\n\t$0\n}\n"
+        }
+    },
+    {
+        "caption": "Preferences: GitGutter Settings - Default",
+        "command": "git_gutter_open_file",
+        "args": {"file": "${packages}/GitGutter/GitGutter.sublime-settings"}
+    },
+    {
+        "caption": "Preferences: GitGutter Settings - User",
+        "command": "git_gutter_open_file",
+        "args": {"file": "${packages}/User/GitGutter.sublime-settings"}
+    },
+    {
+        "caption": "Preferences: GitGutter Key Bindings",
+        "command": "git_gutter_edit_settings",
+        "args": {
+            "base_file": "${packages}/GitGutter/Default (${platform}).sublime-keymap",
+            "default": "// GitGutter Key Bindings - User\n{\n\t$0\n}\n"
+        }
+    },
+    {
+        "caption": "Preferences: GitGutter Key Bindings - Default",
+        "command": "git_gutter_open_file",
+        "args": {"file": "${packages}/GitGutter/Default (${platform}).sublime-keymap"}
+    },
+    {
+        "caption": "Preferences: GitGutter Key Bindings - User",
+        "command": "git_gutter_open_file",
+        "args": {"file": "${packages}/User/Default (${platform}).sublime-keymap"}
+    },
+    {
+        "caption": "Preferences: GitGutter Popup Stylesheet",
+        "command": "git_gutter_edit_settings",
+        "args": {
+            "base_file": "${packages}/GitGutter/gitgutter_popup.css",
+            "default": "/* GitGutter Popup Stylesheet - User */\n\n"
+        }
+    },
+    {
+        "caption": "Preferences: GitGutter Popup Stylesheet - Default",
+        "command": "git_gutter_open_file",
+        "args": {"file": "${packages}/GitGutter/gitgutter_popup.css"}
+    },
+    {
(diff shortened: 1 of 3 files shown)


r-stein:
We may also just remove GitGutter-Edge from the package control and use the prerelease feature.

deathaxe:
I am with you. This would make some things easier and more reliable. I would remove the ${package} thing then and keep the side-by-side setting thing only.

2 questions with it:
1. Am I right, we would need special version tags to mark pre releases?
2. How to efficiently encourage people to move from GitGutter-Edge to Prereleases?

jisaacks:
If y'all want to kill GitGutter-edge, that's fine with me. I might have to be the one to kill it since I added it, not sure actually?

However releases should be cut by myself. I have a script that handles everything for that. But I am happy to cut a release whenever ya'll feel ready, just let me know what version number and change log to set and I'll cut it.

r-stein:
Based on the docs we should move to tag based released anyway:

> #### 4. Pick a Versioning Strategy
> (...)
> *Please note that branch-based releases have been deprecated and no new packages will be accepted that utilize that feature. Existing packages are encouraged to update to using tags.*

[source](https://packagecontrol.io/docs/submitting_a_package)

It shold not be that hard to join the packages by adding GitGutter-Edge to the `previous_names` array and removing the old entry. From [Renaming a Package](https://packagecontrol.io/docs/renaming_a_package)

jisaacks:
> It shold not be that hard to join the packages by adding GitGutter-Edge to the previous_names array and removing the old entry. From Renaming a Package

That makes sense, is that something you can do or am I the only one with permissions to do so?

r-stein:
Anyone can create a PR, but you should add a message there, that you approve it.

deathaxe:
I confirm @jisaacks is the only one to cut releases. I would never ever do something this on my own. I asked the question about tagging only because it is not clear to me how pre-releases are handled by package control and what is required to support them in general. I recently added `PackageDev` to the list of pre_releases but this is just the user side. I didn't check the devs one.

One idea about cutting releases and changelogs:

I found GitSavvy having an interesting function called `git: generate changelog`. It uses the first word of the first commit message text separated by : to categorize changes. They use (fix,  enhancement, feature) as categories. I find this function quite useful to quickly generate a changelog from the existing commits, which then only needs little polish if all commit messages are formated correctly. Is it something we might use, too, to make the process of cutting a release a bit easier?

jisaacks:
> it is not clear to me how pre-releases are handled by package control 

I believe it is whether or not the release is marked as one when you create a release via github. I am doing it via the github API. you can check it out here: https://github.com/jisaacks/GitGutter/blob/master/release_messages/src/release.sh#L3-L5

> I found GitSavvy ... Is it something we might use, too, to make the process of cutting a release a bit easier?

If I can easily add it to my release tool that I built, I am fine with it.

deathaxe:
I don't know if you can or if it helps you to make things easier. It is a normal sublime text command which is called from command pallet. It creates a new view and adds all commit messages categorized.

Example:

```
Changes since 1.4.0:

  Other:
   - Fix release notes typo (#299)
   - Fix typo in 1.4.0 release message
   - Change the way the popup css files are loaded
...
  Fix:
   - Show the correct branch or tag name in the 'compare against' status message.

  Enhancement:
   - Make the status message somewhat smarter.
   - Use jinja2 template engine to render status bar messages.

  Feature:
   - Add 'Compare against file commit' and enable filtering by commit id (Issue #281)
...
```

Every message not beginning with `<word>:` is add to the `Other:` node.

FichteFoll:
Package Control does not interact with Github's Releases, only with tags. Providing a pre-release works according to http://semver.org/, i.e. by adding a `-pre-release` suffix to a potential future release. See also https://github.com/SublimeText/PackageDev/tags.

You should include an update message in your next release indicating that previous GitGutter-Edge users should add `GitGutter` to their `install_prerelease` setting array in PC's setting file. See also https://packagecontrol.io/docs/settings.

deathaxe:
@FichteFoll: Thanks very much for clarification.

jisaacks:
@deathaxe I am fine using that release strategy but how does it know which commits are new since the last release? We'd also have to start making our commit messages more uniform.

Also thanks for the info @FichteFoll

deathaxe:
### Rebase and remove ${package} support

As we want to support one package only, it might not be required to support an envirionment variable ${package}. So I removed it and simplified the PR a bit to only keep side-by-side settings support.

### Releas strategy
Discussing in #341 maybe better?

PR #360, opened 2017-01-02
Update README.md: 3124 is no longer a dev build

Arnie97:


r-stein:
Thanks for updating that info.

PR #361, opened 2017-01-04
Fix Issue #256: Add support for custom icon packages.

deathaxe:
A new setting `theme` provides a value of `<Name>.gitgutter-theme` which is an
empty file passed to `sublime.find_resources(...)` to find the folder containing
the gutter icons to show for the events and the gitgutter_popup.css to use for
the diff popup. So theme developers can add custom icons and popup styles for
GitGutter to their package. The handling is derived from Sublime Text's default
`theme` handling to keep the setting as easy as possible, so user doesn't need
to care about paths.

As Sublime Text 2 does not support for `sublime.find_resources(...)` only
built-in themes are allowed then.

The commit contains two built-in themes `Default` and `Default HIDPI` to support
high resolution displays for older versions of Sublime Text.

Most themes already provide name@2x.png files to support high resolution
displays without any special setting being required. Sublime Text is expected to
use those files incase of DPI scale > 1. So GitGutter may or may not provide the
larger icons with this naming in the `Default` theme as well incase Sublime Text
will support them in the future.

Note: This provides an alternitive solution for PR #347


git_gutter_popup.py
```diff
@@ -195,21 +195,24 @@ def is_button_enabled(k):
...
         )
         content = button_line
-    css = ''
+
     if _MD_POPUPS_USE_WRAPPER_CLASS:
         wrapper_class = ".git-gutter"
     else:
         wrapper_class = ""
 
-    # load the css files
```
r-stein: This searched all `gitgutter_popup.css` files before. Now you changed it to only search those defined in a setting. I don't think that's an improvement.
Also it is performance wise better to create a (mutable) string array and join it than to add each single string, which creates a new immutable string.

deathaxe: The idea is to encourage theme devs to add a `git_gutter_popup.css` to their package to modify the look of the diff popup. As a result there may be many packages containing this file. Merging all of those files would lead to unpredictable results.

By now the css will most likely exist in GitGutter package and maybe in User package, too. So you have 2 files to merge, where user settings overwrite the settings from our default css.

This is what I intend to keep but without restricting the location of the css file.

So:
1. Load default file from GitGutter package
2. Merge in theme specific file to overwrite GitGutter defaults
3. Merge in user specific file to overwrite 1 and 2.

If creating string tuble and merge by join is better, I am not refuced to change this again. ;-) even though I think this makes a big difference with 3 items.

deathaxe: Btw.: Replacing `${package}` by the real folder name is used by `Annaconda` package, too, to make those commands work, even if the package is locally renamed. The only difference is they use `$package_folder` as variable.

Issue #290, opened 2016-07-13
No support for submodules


oferei:
I have a git submodule within my git project. When I open a submodule file in Sublime I see plus signs beside all the lines, as if the entire content is newly added.

oferei:
and thanks for this plugin!  ^__^

oferei:
You assume right - git diff shows no differences.
I'm using macOS Sierra (v10.12.1) and git version 2.9.3 (Apple Git-75).

rchl:
Is it reproducible for you on any public repo, so that I could try it locally? It works with the submodule for me in my repo.
Tried using both git 2.9.1 and git 2.7.4 (Appplie Git-66).

oferei:
Created a brand new project on GitHub, but could not reproduce - worked as expected.
Will keep investigating.

deathaxe:
This issue may be caused by some versions of git itself. Some git commands randomly fail, if git is not called from within the working tree, but this is exactly what GitGutter does.

I prepared a fix to run all git commands from within the working tree, which is published as soon as all required PRs are accepted.

deathaxe:
Should work with GitGutter 1.5.0. Same as with work-tree support. Please check.

oferei:
It works with 1.5.0!    (^O^)／

Issue #248, opened 2015-08-29
Does not support Worktree from Git 2.5+


nj:
When editing code in a [worktree](https://github.com/blog/2042-git-2-5-including-multiple-worktrees-and-triangular-workflows), all lines appears as "new" (`+`'s) and not the actual changed lines within a file.

`git status` shows the expected value, so it's only wrong within Sublime.

Tested on `Sublime v3 build 3083` and `GitGutter v1.2.1`

jisaacks:
Thanks for pointing this out.

> Please note that the worktree feature is experimental. It may have some bugs, and its interface may change

I think I may wait for it leave it's experimental phase and lock in it's interface before adding code to support it.

vsimon:
The worktree feature has rounded out since the git 2.5 days. Would this be a better time to support this? (I've found the feature itself to be super useful).

deathaxe:
I prepared a fix to run all git commands from within the working tree, which is published as soon as all required PRs are accepted.

This implicitly enables work-tree support.

deathaxe:
Should work with GitGutter 1.5.0 Please check

deathaxe:
As submodules work work-trees also do. It's the same mechanism.

Issue #239, opened 2015-07-14
Respect the case-sensitivity setting in git's config


alextegelid:
When I rename a file from `filename.php` to `Filename.php` GitGutter will show every row in that file as added (thinking it's a new file). `git status`, on the other hand, reports a clean working tree since my global settings file says `core.ignorecase=true`

deathaxe:
This is actually a bug of git so far. as calling `git show Filename.php` works case insensitive but `git show HEAD:Filename.php` does not.

PR #364, opened 2017-01-28
Move the UI updates to the main thread

### LINKED ISSUE #340: Noticeable flicker/delay in icons when working with files

r-stein:
This fixes #340 by moving the UI updates to the main thread.

If have tested it on ST3 and ST2. We may just issue a release soon, because many users may be anoyed by the flicker.

### DIFF
--- a/git_gutter_show_diff.py
+++ b/git_gutter_show_diff.py
@@ -59,7 +59,8 @@ class GitGutterShowDiff(object):
         # update the if lines changed
         elif self.diff_results is None or self.diff_results != contents:
             self.diff_results = contents
-            self._update_ui(contents)
+            # move to the main thread and update the ui
+            sublime.set_timeout(lambda: self._update_ui(contents), 0)
 
     def _update_ui(self, contents):
         inserted, modified, deleted = contents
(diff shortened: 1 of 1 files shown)


deathaxe:
I finished working on this issue already a month ago. The pending PR #354 and #356 prevent me from publishing the solution as it is based on them and requires some of the changes introduced by them.

As you made this suggestion earlier this was my starting point, too with following result:

Moving all of this to the main thread will have two effects.
1. Step 2 (see below) may finish a bit quicker as inter process communication with plugin_host.exe is not required.
2. User interface will lack again if files and changes grow.


The main reason for flickering is the time required to calculate the new regions. 
Current *steps* are:
1. clear all regions (icons) _-> icons removed_
2. convert the git diff result to regions
3. add new regions. _-> icons added_

Step 2 easily takes 100ms to 10s on larger files (ca. 100k lines) right now. This is what the user sees as flicker if the edited file becomes a bit larger with many diffs int it. The reason for the delay is the very heavy use of Sublime API functions which are called severel hundreds or thoughand times. Sublime API functions are "very" slow and working on the text buffer directly within an asynchronous function is a no go at all as the text buffer may change in the mean while.

Therefore I had a look on SublimeLinter which parses text in a dedicated thread, too. They once read the whole text on the beginning of the whole process of parsing by `view.substr()` and work with this text to do their calculations. This avoids most issues if the underlying text changes a bit and is much faster. By using the same strategy I was able to reduce the time needed for step 2 to 1..10ms for most files. Parsing 100k lines needs 150ms (10s before!!!) only.

So I would be pleased to get at least the two PR #354 and #356 approved to be able to publish my work on Issue #340 so we can discuss it.

This work includes fixes for issues #239, #248, #290, two, and includes some patches to reduce the ammount of git file updates to decrease latency a bit.

r-stein:
Close in favor of #372

PR #365, opened 2017-02-02
Fix: Run all git commands from within working tree

deathaxe:
This is the last requirement for the fix of Issue #340, but I think it is worth a seperate PR.

#### Issue #239:
This commit does not yet fix case sensitive issue with  `git show -- REF:file` but helps git to respect config at all.

#### Issue #248 
A linked work-tree's root directory contains a file called '.git' which contains the path to the native repository. By code GitGutter checks for existent '.git' file but it seems not to work reliably on some platforms paired with `--git-dir` or `--work-tree` arguments.

#### Issue #290 
Same as with #248.

### Description

GitGutter currently uses `--git-dir` and `--work-tree` arguments to pass the required information about the repository and its work-tree to git, but it seems git doesn't handle those arguments reliably. Commands running smoothly executed from shell in context of the working directory, fail when called by
GitGutter. Either git output is empty or at least settings or `.gitattributes` are ignored. The behavior changes from call to call. One example is `git status`, which fails the first time when called with `--git-dir` and `--work-tree` and current directory being something outside the working tree. 
(Tested with Windows x64 / git 2.11.0)

To avoid any trouble with git's commands/features the work-tree path is passed to `POpen` as `cwd` to perform a chdir to it before spawning git process. This makes `--git-dir` and `--work-tree` arguments obsolete as git finds everything it needs to be happy on its own. This is the way other plugins handle git
execution, too.

The `git_handler.work_tree()` method replaces `on_disk()` and checks, if the current view's file is part of a valid git repository. One of the parent directories must therefore contain a folder or file called `.git`, but the file path itself must not contain `.git` as this indicates a file within the database or something like COMMIT_EDITMSG file which does not belong to the work-tree and therefore must not be handled by GitGutter.

The `git_dir` attribute is no longer needed and therefore removed as `_git_tree` is used as primary key for all required settings.

In the end those changes made `git_helper` module obsolete.

GitGutter detects renamed files now.
1. If the file is renamed within a work-tree its state changes to untracked.
2. If it is no longer part of a valid work-tree the state is completely cleared.
3. By  moving it back into its original position the state is restored.

Same works by removing and restoring the `.git` directory. Means, if you call `git init` in the open file's directory, GitGutter will update correctly.


deathaxe:
@r-stein : Are you willing to review and discuss the changes introduced by this PR?

I found a way to simplify the process of determining whether a view shows a valid file within a working tree and find this PR ready to merge.

Would be nice, if the changes were double checked by anyone else to ensure I did not miss anything.

Issue #366, opened 2017-02-02
Unable to locate module [jinja2]


magnetikonline:
Hey JD,

   love your work on this plugin - one of my "must haves" when using ST3.

But have noted since release https://github.com/jisaacks/GitGutter/commit/2cc9a1a9e780e9280480c6c17d80c214af5c1245 I'm getting a Python module error as follows:

```startup, version: 3126 osx x64 channel: stable
executable: /Applications/Sublime Text.app/Contents/MacOS/Sublime Text
working dir: /
packages path: /Users/magnetik/Library/Application Support/Sublime Text 3/Packages
state path: /Users/magnetik/Library/Application Support/Sublime Text 3/Local
zip path: /Applications/Sublime Text.app/Contents/MacOS/Packages
zip path: /Users/magnetik/Library/Application Support/Sublime Text 3/Installed Packages
ignored_packages: ["Vintage"]
pre session restore time: 0.193298
using gpu buffer for window
first paint time: 0.247873
using gpu buffer for window
using gpu buffer for window
startup time: 0.302103
reloading plugin Default.auto_indent_tag
reloading plugin Default.block
reloading plugin Default.comment
reloading plugin Default.convert_syntax
reloading plugin Default.copy_path
reloading plugin Default.delete_word
reloading plugin Default.detect_indentation
reloading plugin Default.duplicate_line
reloading plugin Default.echo
reloading plugin Default.exec
reloading plugin Default.fold
reloading plugin Default.font
reloading plugin Default.goto_line
reloading plugin Default.history_list
reloading plugin Default.indentation
reloading plugin Default.install_package_control
reloading plugin Default.kill_ring
reloading plugin Default.mark
reloading plugin Default.new_templates
reloading plugin Default.open_context_url
reloading plugin Default.open_in_browser
reloading plugin Default.pane
reloading plugin Default.paragraph
reloading plugin Default.paste_from_history
reloading plugin Default.profile
reloading plugin Default.quick_panel
reloading plugin Default.run_syntax_tests
reloading plugin Default.save_on_focus_lost
reloading plugin Default.scroll
reloading plugin Default.set_unsaved_view_name
reloading plugin Default.settings
reloading plugin Default.show_scope_name
reloading plugin Default.side_bar
reloading plugin Default.sort
reloading plugin Default.swap_line
reloading plugin Default.switch_file
reloading plugin Default.symbol
reloading plugin Default.transform
reloading plugin Default.transpose
[...]

PR #367, opened 2017-02-03
Fix: Avoid exception due to missing dependencies. (Issue #366)

deathaxe:
If Sublime Text starts the very first time with GitGutter installed some
dependencies may not be satisfied yet, which results in an uncaught exception.

To avoid issues with the whole plugin loading process the depending functions
are disabled, if dependencies are not satisfied.

This commit disables the status bar text, if jinja2 library is missing.

### DIFF
--- a/git_gutter_show_diff.py
+++ b/git_gutter_show_diff.py
@@ -1,6 +1,13 @@
 import os
-from jinja2.environment import Template as Jinja2Template
 import sublime
+
+try:
+    # avoid exceptions if dependency is not yet satisfied
+    from jinja2.environment import Template as Jinja2Template
+    _HAVE_JINJA2 = True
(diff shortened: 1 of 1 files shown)


git_gutter_show_diff.py
```diff
@@ -89,8 +96,8 @@ def _update_status(self, file_state, contents):
             file_state - the git state of the open file.
             contents   - a tuble of ([inserted], [modified], [deleted]) lines
         """
-        template = settings.get('status_bar_text') \
-            if settings.get('show_status_bar_text') else None
+        template = settings.get('status_bar_text') if _HAVE_JINJA2 and \
```
r-stein: Please use parens instead of `\` for line continuation.

PR #368, opened 2017-02-04
Readd dependencies to >=3080 entry

r-stein:
The dependencies should be duplicated.

> The most-specific selectors will be used, and all other ignored. This
> means that some dependencies will be duplicated under a specific
> platform and under *.

https://github.com/wbond/package_control/blob/master/example-dependencies.json

### DIFF
--- a/dependencies.json
+++ b/dependencies.json
@@ -7,7 +7,9 @@
         ">=3080": [
             "pygments",
             "python-markdown",
-            "mdpopups"
+            "mdpopups",
+            "python-jinja2",
+            "markupsafe"
         ]
     }
 }


deathaxe:
The review thing is terrible here!!!

We need this:

```js
{
    "*": {
        "<3080": [
            "markupsafe",
            "python-jinja2"
        ],
        ">=3080": [
            "markupsafe",
            "python-jinja2",
            "pygments",
            "python-markdown",
            "mdpopups"
        ]
    }
}
```

deathaxe:
But I encountered an issue with installing dependencies using ST2 on Linux. Package Control fails with an parse error. ST2 Windows works.

see: https://github.com/wbond/package_control/issues/1213

r-stein:
`*` is less specific, but `>=3080` does not match. Even in the example `*` is used as a fallback if the more specific one does not match.

deathaxe:
I tried on Linux ST2 some mins ago with same solution as yours. ST2 simply ignored the packages. The only thing make it work was to use "<3080".

deathaxe:
Just double checked. Seems to work with "*", too. I caught myself somewhere.

PR #369, opened 2017-02-05
Refactor README

deathaxe:
As GitGutter-Edge is removed from Package Control I started to have a look onto the README.md to update the information how to receive pre-releases in the future.

Doing so I found several possible improvements we might discuss.

Things suggested for improvement:

1. A List of the main features is added to the brief description on the beginning as GitGutter provides much more then gutter icons now.
2. Dependencies
   a) A list of required libraries is added to *Installation* so people might help themselfs on manual setups.
   b) The required dependency is displayed in the description of each *Feature* and *Setting*
3. Eual formatting templates for
   a) key bindings
   b) step by step descriptions
   c) settings paragraphs
4. Try to use the same words and phrases everywhere.
5. Use passive time form to describe features and settings.
6. Add some icons to create some eye catchers and break overwhelming amount of text.

The whole document is reformated to satisfy the linting rules proposed by PR #339.


README.md
```diff
@@ -1,51 +1,139 @@
...
+
+## 🚀 Advanced Features
 
 ### Diff Popup
 
+The diff popup appears by hovering the mouse over the gutter changes on Sublime
+Text 3 or can be called from the command palette by `GitGutter: Show Diff
+Popup` or via a key binding.
+
+ⓘ *popups require Sublime Text 3 Build 3070+, [python-jinja2][python-jinja2]*
```
r-stein: It requires all dependencies. However since they are automatically installed, we should just omit that. People, who don't use Package Control are a small minority and they are listed above anyway.

README.md
```diff
@@ -58,104 +146,215 @@ The most common reason for the icons to not show up is likely a problem with Git
...
+
+Default settings should not be modified, as they are overwritten when GitGutter
+updates. Copy the relevant settings into GitGutter's user settings file instead.
 
-Default settings should not be modified, as they are overwritten when GitGutter updates. Instead, you should copy the relevant settings into GitGutter's user settings file.
 
 #### Debounce Delay
-Delays update of gutter icons by the following amount (in milliseconds). Useful for performance issues. Default 1000 (1 second).
+
+> `"debounce_delay": 1000`
```
r-stein: Using citations `>` for this does not look good.

README.md
```diff
@@ -58,104 +146,215 @@ The most common reason for the icons to not show up is likely a problem with Git
...
-In a POSIX environment you can run `which git` to find the path to git if it is in your path.  On Windows, you can use `where git` to do the equivalent.
 
 #### Protected Regions
-Is GitGutter blocking SublimeLinter or other icons? You can prevent this by adding which regions you would like GitGutter to not override:
-```javascript
-"protected_regions": ["region", "names"]
-```
-You will need to figure out the names of the regions you are trying to protect.
 
+> ```
```
r-stein: This should just be a block not a cited block.

README.md
```diff
@@ -58,104 +146,215 @@ The most common reason for the icons to not show up is likely a problem with Git
...
+>   "In {{repo}} on {{branch}}",
+>   "{% if compare != 'HEAD' %}, Comparing against {{compare}}{% endif %}",
+>   ", File is {{state}}",
+>   "{% if deleted != 0 %}, {{deleted}}-{% endif %}",
+>   "{% if inserted != 0 %}, {{inserted}}+{% endif %}",
+>   "{% if modified != 0 %}, {{modified}}≠{% endif %}"
+> ]
+> ```
+
+ⓘ *requires [python-jinja2][python-jinja2]*
```
r-stein: I would just remove that line, because it is more distracting than helpful.

README.md
```diff
@@ -58,104 +146,215 @@ The most common reason for the icons to not show up is likely a problem with Git
...
+
+The *Status Bar Text* is rendered using a fully customizable template from
+`status_bar_text`. The setting is organized as an array of strings for better
+readability. It is joined and then passed to [python-jinja2][python-jinja2].
```
r-stein: We should link to [jinja2](http://jinja.pocoo.org/docs/) not to the dependency and describe to "how to use it" not "how it is handled in the background".

README.md
deathaxe: Sure? I somehow liked the gray bar on the left.

Btw. It quotes the default value from settings file  :-)

README.md
deathaxe: All blocks quoting default values from settings file should have the same style.

So we need to decide whether to use `>` for all settings which draws them with normal font or use the codeblock to draw them in fixed font width to state it as a special setting.

I found combinding both would be good, but can revert to one of them.

README.md
deathaxe: Sounds good.

README.md
deathaxe: ok.

README.md
deathaxe: With a second look on it, you might be right.

I'll change it to code block but without highlighting.

README.md
```diff
@@ -1,51 +1,138 @@
...
+it automatically installs required [dependencies](#dependencies) and keeps all
+packages up to date.
+
+1. [Install][3] Package Control if you haven't yet.
+2. Open the command palette (<kbd>Ctrl+Shift+P</kbd> for Windows/Linux,
+   <kbd>Cmd+Shift+P</kbd> for Mac)
+3. Search for *Package Control: Install Package* and hit <kbd>Enter</kbd>.
+4. Type `GitGutter` and press <kbd>Enter</kbd> to install it.
+5. Open the command palette again.
+6. Search for *Package Control: Satisfy Dependencies* to install required
```
r-stein: This is not necessary it will automatically check the dependencies. You usually only need that command to check for updates on the dependencies.

r-stein:
A general note: I would prefer `[title](line)` over `[title][1] ... [1]:<link>`

deathaxe:
I used this construct for two reasons:

1. urls like to break the 80 chars limit and therefore doen't looks so good in the middle of the text. Having all links together at the bottom of the document keeps all the long lines in one place at least.
2. Some urls are used in several places. So it's like a variable definition done once used many.

deathaxe:
The 80 char limit was applied due to the coala PR and the complains about my reply not to apply this rule.

1. With MarkdownEditing package or text wrapping enabled, the text looks good in editor anyway.

2. The fact with changing several lines due to one word is a real drawback with git history in mind. A small change might look like a big one then.

3. When creating the most recent commit I destroyed some links several times by `Alt+Q` wrapping. I found this a bit dangerous. Maybe it is a good idea to just replace the numbered foot notes by names. I agree with numbers might not be a good tagging strategy.

I am not sure about what is the best "coding style" for markdown. I didn't use it too often in the past, even though I find it a quite cool thing.

r-stein:
In LaTeX it is usually a good rule to have on line per sentence. In markdown this doesn't seem to bad either. A problem could be that some markdown to html converter adds a `<br>` at newlines (github doesn't do this at READMEs, not sure about the Package Control page).

Markdown is mostly for writing stuff readable with very limited structures. Selecting how to write, i.e. `*italic*` or `_italic_` is mostly a personal preference. Same holds for the links.
One could argue for the inline links, that they are similar to the resulting document, because you can open them at the place without searching the reference.
On the other side the reference style mimics the typical paper structure, where you have citation like `[1]` in the text and a bibliography at the end of the document.
However I don't care to much about the link style, but prefer the inline links.

deathaxe:
I found [daringfireball.net][1] a good homepage with a set of helpful tips and rules.

The most important sentence about styling markdown in this document is:

> A Markdown-formatted document should be publishable as-is, as plain text,
> without looking like it’s been marked up with tags or formatting instructions.

## Line length

Only very few sources even mension recommended line lengths at all, but _Markdown_ is activly said not converting linebreaks to `<br>`. Even though this does not work for GitHub's post editor here.

With RFC plain text documents in mind a fixed line length might be a good idea.

I tried your one sentence one line rule, which basicly sounds good, but it has one big disadvantage. Depending on length of a sentence a line can be very short or very long. Turning on auto line wrapping then may cause a sentence continue on a second line with only one or two words, followed by a line break and another sentence. This looks ugly. Without wrapping long lines are hard to read. This is not what I would like to look the text representation of the document.

## Formatting text

Some pages recommend different styles for italic and bold as follows
- \_italic\_ should be marked by underscore
- \*\*bold\*\* by astarix

- lists may be inteded by up to 3 characters

GitHubs [style guide][2] states about tables:

- tables rows should/need not start or end with |

## Links

The [daringfireball.net][1] page makes follwoing recommendation about links.

> The point of reference-style links is not that they’re easier to write. The point is that with reference-style links, your document source is vastly more readable. Compare the above examples: using reference-style links, the paragraph itself is only 81 characters long; with inline-style links, it’s 176 characters; and as raw HTML, it’s 234 characters. In the raw HTML, there’s more markup than there is text.

> With Markdown’s reference-style links, a source document much more closely resembles the final output, as rendered in a browser. By allowing you to move the markup-related metadata out of the paragraph, you can add links without interrupting the narrative flow of your prose.

This is what I naturally was thinking about, when adding those references. They disturb text reading much less, then direct links. And IMHO it avoids to accidently wrap an url and draw it invalid.


[1]: http://daringfireball.net/projects/markdown/syntax
[2]: https://guides.github.com/pdfs/markdown-cheatsheet-online.pdf

deathaxe:
@jisaacks / @r-stein : I just saw the GitGutter's readme not being rendered correctly on packagecontrol.io due to some line break issues.

To fix it and tell people about pre-releases, we might finish this PR, soon.

Do you have any change requests left open or would you approve?

jrappen:
@deathaxe I strongly agree with @r-stein , markdown documents are not source code. Breaking lines after each sentence within any given paragraph is highly preferred for git repositories compared to an 80 chars line break, if you **really really really** need any line break at all. Maintainers would simply give up on editing the documentation if you enforce an 80 chars because quite frankly, highly repetitive tasks are a [pita](http://www.urbandictionary.com/define.php?term=pita).

deathaxe:
I don't enforce anything. I did a suggestion and posted some reasons for my decision. 

When I complained about PR #366 which suggested those line-breaks, I was shit stormed last autumn so found it a common rule to do and even tried to find some docs about it. In fact I hardly found any statement about line breaks in markdown at all. The only usable statement was found on daringfireball.net

I use ST's wrapping feature <kbd>Alt+Q</kbd> to handle line breaks efficiently semi-automatically, so it wasn't so repetitive and expensive. It is dangerous sometimes, because wrapping urls breaks them.

Many editors do not automatically wrap Markdown, so the rule ...

> A Markdown-formatted document should be publishable as-is, as plain text, without looking like it’s been marked up with tags or formatting instructions.

... made sense to me.

If line breaks are unwanted, I would suggest to have one line per paragraph. I tried with one line per sentence as suggested by @r-stein, but it causes ugly line breaks with auto wrapping enabled in an editor sometimes.

A real disadvantage of line breaks as @r-stein stated is `diff` at all. Any small change which requires a line break makes the whole paragraph look changed.

deathaxe:
Removed all hard line breaks and footnotes as I am with you in the end that limiting line length for GitHub readmes has too many disadvantages.

r-stein:
You can see the README in rendered format here: https://github.com/deathaxe/GitGutter/tree/pr/readme

jisaacks:
LGTM 👍

r-stein:
@jisaacks You can click on the upper tab at *Files Changed* and then top left at *Review Changes* and then Approve it. I think we should generally use that (relative new) technique instead of dropping comments.

jisaacks:
@r-stein thanks for pointing that out :)

PR #370, opened 2017-02-05
Fix: Fall back to simple status bar text if jinja2 is not available.

deathaxe:
The recently added support for status bar text rendering using powerful python-jinja2 template engine results in no status bar messages being displayed at all, if jinja2 is not available for any reason.

So GitGutter might fallback to a simple message mode, if python-jinja was not found.

It becomes an optional dependency then which enables users to modify the way status bar messages ar rendered.

Please vote for adding this feature or not.

### DIFF
--- a/git_gutter_show_diff.py
+++ b/git_gutter_show_diff.py
@@ -3,10 +3,9 @@ import sublime
 
 try:
     # avoid exceptions if dependency is not yet satisfied
-    from jinja2.environment import Template as Jinja2Template
+    import jinja2.environment
     _HAVE_JINJA2 = True
 except:
-    print('GitGutter: Missing jinja2 dependency!')
     _HAVE_JINJA2 = False
 
 try:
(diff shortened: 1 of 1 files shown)


git_gutter_show_diff.py
```diff
@@ -96,22 +95,43 @@ def _update_status(self, file_state, contents):
...
-        if template:
-            def set_status(branch_name):
-                inserted, modified, deleted = contents
+        if not settings.get('show_status_bar_text', False):
+            self.view.set_status('00_git_gutter', '')
+            return
+
+        def set_status(branch_name):
+            inserted, modified, deleted = contents
+            template = settings.get('status_bar_text') \
```
r-stein: Please use parens:

``` python
template = (
    settings.get('status_bar_text')
    if _HAVE_JINJA2 else None
)
```

> The preferred way of wrapping long lines is by using Python's implied line continuation inside parentheses, brackets and braces. Long lines can be broken over multiple lines by wrapping expressions in parentheses. These should be used in preference to using a backslash for line continuation. 

[pep8](https://www.python.org/dev/peps/pep-0008/#maximum-line-length)

git_gutter_show_diff.py
```diff
@@ -96,22 +95,43 @@ def _update_status(self, file_state, contents):
             file_state - the git state of the open file.
             contents   - a tuble of ([inserted], [modified], [deleted]) lines
         """
-        template = settings.get('status_bar_text') if _HAVE_JINJA2 and \
-            settings.get('show_status_bar_text') else None
-        if template:
-            def set_status(branch_name):
-                inserted, modified, deleted = contents
+        if not settings.get('show_status_bar_text', False):
+            self.view.set_status('00_git_gutter', '')
```
r-stein: We may better use `self.view.erase_status('00_git_gutter')`

git_gutter_show_diff.py
```diff
@@ -96,22 +95,43 @@ def _update_status(self, file_state, contents):
...
-        else:
-            self.view.set_status('00_git_gutter', '')
+            else:
+                # Render hardcoded text if jinja is not available.
+                parts = []
+                parts.append('On %s' % branch_name)
+                compare = self.git_handler.format_compare_against()
+                if compare != 'HEAD':
+                    parts.append('Comparing against %s' % compare)
+                parts.append('File %s' % file_state)
```
r-stein: I would avoid that in the default mode. I don't think the file state is so interesting, because you know that anyway.

deathaxe: All 4 lines or the file state only?

git_gutter_show_diff.py
deathaxe: Interesting feedback :-) as I am looking for some style guides, too, at the moment as some situations are not totally clear to me how to solve correctly. I am always willing to make code as clean as  possible.

git_gutter_show_diff.py
r-stein: Just the file state. The compare against may be useful unless it is HEAD. The branch may also be helpful.

git_gutter_show_diff.py
deathaxe: But why does SublimeLinter not complain about ?! Hmm.

Issue #74, opened 2013-03-18
Problem with smudge/clean filters


fd:
Smudge/clean filters are not applied by `git show` which causes GitGutter to detect ghost-changes.

kofalt:
Hello @jisaacks!  I just ran into this today and I can help lend some context.

Git filters allow you to modify files when checking out or committing. You can do a couple cool things with this: store & version an encrypted file alongside your normal source code, for example. I think the Windows Git client uses filters to solve compatibility problems with line endings as well.

Docs are [here](https://git-scm.com/book/en/v2/Customizing-Git-Git-Attributes#Keyword-Expansion) but the pictures are pretty great:

![smudge](https://cloud.githubusercontent.com/assets/265071/13125849/49e1332a-d58d-11e5-975f-c3f2ec840eb0.png)

![clean](https://cloud.githubusercontent.com/assets/265071/13125853/4b76f116-d58d-11e5-9e4f-4c79ebc63092.png)

Here's a popular python library, edited locally as tabs, without messing up the diff:

```bash
# Grab an example repo
git clone https://github.com/kennethreitz/requests.git && cd requests

# Create a filter called "tabspace"
git config filter.tabspace.smudge 'unexpand --tabs=4 --first-only'
git config filter.tabspace.clean  'expand   --tabs=4 --initial'

# Tell git to run the filter on all python files
echo '*.py  filter=tabspace' >> .git/info/attributes

# Bit silly: delete all tracked files & checkout again to trigger the filter
git rm -rf -q .
git reset -q .
git checkout .

# Have a look!
sublime_text requests/cookies.py
```

You'll see that the `smudge` filter has been applied and the file is indented with tabs, unlike its [upstream variant](https://github.com/kennethreitz/requests/blob/master/requests/cookies.py). Meanwhile, `git diff` is empty, because the `clean` filter applies before diffing or comitting.

You'll also see that GitGutter thinks each line changed by a filter has a diff status:

![gutter](https://cloud.githubusercontent.com/assets/265071/13126704/275ac8e4-d591-11e5-8c23-57edf4f5c7c9.png)

I think this is what @fd was getting at - that GitGutter's method for checking the diff ([here?](https://github.com/jisaacks/GitGutter/blob/6f3d9783aec01ac5efc8456bfe78c5fcdc6dec0c/git_gutter_handler.py#L154)) doesn't account for git filters, which leads to erroneous reporting.

[...]

Issue #263, opened 2015-12-31
Change marks not clearly visible on the minimap


azdavis:
<img width="1280" alt="screen shot 2015-12-30 at 8 27 55 pm" src="https://cloud.githubusercontent.com/assets/6496454/12059962/dcb3e272-af33-11e5-9451-26fc6cf790c7.png">

I verified the default settings have `"show_in_minimap": true` and even added this to my own user settings. No dice.

| thing | version |
| :-- | :-- |
| GitGutter | 1.2.2 |
| Sublime Text | 3 (dev build 3095) |
| OS X | 10.11.2 (build 15C50) |

AviBueno:
Same issue on Win 8.1, Git Gutter 1.2.2, Sublime Text 3 stable build v3083.

jisaacks:
@azdavis @AviBueno Is it not working at all or just under certain circumstances?

azdavis:
I've done some experimentation.

Adding lines with text, it seems as though the green bar is actually behind the text.
<img width="162" alt="img_1" src="https://cloud.githubusercontent.com/assets/6496454/12126121/4c950412-b3bd-11e5-9a39-a53787a7c289.png">
This problem, as one might expect, does not manifest when the added lines have no text of their own.
<img width="171" alt="img_2" src="https://cloud.githubusercontent.com/assets/6496454/12126148/6cb320d0-b3bd-11e5-8092-426508c0c157.png">
I thought maybe if I added or modified lines with indentation, this problem would be avoided. No dice.
<img width="1280" alt="img_3" src="https://cloud.githubusercontent.com/assets/6496454/12126166/855698a6-b3bd-11e5-8e8a-c92db0df7c46.png">
But then I deselected that which I had indented…
<img width="1280" alt="img_4" src="https://cloud.githubusercontent.com/assets/6496454/12126176/91a99f5e-b3bd-11e5-9a28-34c4e2d4a58a.png">
Odd indeed.

AviBueno:
Sorry, I have GitGutter installed for a long time but only yesterday I discovered it can show up on the minimap. This was after I saw the Atom equivalent on a friend's machine. So I looked at the minimap but didn't see anything.
Following @azdavis reply I only now realized it's a 1 pixel thingy (was expecting a fully colored line).

I tested now and I guess it works, although it's almost impossible to see any indication.
[...]

PR #372, opened 2017-02-12
A set of patches to evolve to a possible pre-release 1.5.0 pre.1

deathaxe:
I'd like to propose fixes for issues #74, #263, #340 and several enhancements to reduce disk usage and number of calls to `git show` / `git archive`.

I put all commits into this PR I would find worth to be included in the next release as they fix some of the most annoying issues.

For details about the changes please have a look onto the commit messages itself.

I hope to find some willing people to check the comprehensive changes thoroughly. I work with them for a while now and found them quite stable during the last days and weeks.


git_gutter_settings.py
```diff
@@ -178,6 +172,15 @@ def set_compare_against(self, git_tree, new_compare_against):
...
         self._compare_against_mapping[git_tree] = new_compare_against
 
+    def get_show_in_minimap(self, view):
+        """Read 'show_in_minimap' setting and apply appropiate limits."""
+        if not ST3:
+            return 0
+        width = view.settings().get('git_gutter_show_in_minimap')
+        if width is None:
+            width = self.get('show_in_minimap', 1)
+        return width if 0 <= width <= 9 else 100000
```
r-stein: You must also handle the values `True` and `False`, because they were valid and hence still in some settings files.

deathaxe: I checked by setting to true and false in may user settings with success.

`False` means 0 and therefore hides gutter icons from minimap.
`True` means 1 and shows gutter icons with default width of 1px as before.

Did you encounter any functional issues with it?

kofalt:
I don't think I posted here, but I can confirm that this branch fixed #74 for me :+1: 

If you need any further testing, let me know!

deathaxe:
As this PR changes quite much, I would be pleased to any feedback about possible bugs in general. Are the gutter icons always updated as expected? Are there situations a modification in a text file or external status change of the file (stage, commit) is not recognized? Do you encounter any exceptions? Things like that would help. If no such strange things can be found and no one complains about code formating, I'd like to merge it to the master branch and ask to publish it as pre-release somewhat later, so more people can test it before final release.

kofalt:
To answer your questions:

* Are the gutter icons always updated as expected?

Yes, as far as I could tell. I even did some work over SSHFS and it was fine with a higher-latency file system. 

* Are there situations a modification in a text file or external status change of the file (stage, commit) is not recognized?

I could not find any situation where that happened - it does seem like there is a delay on initial file load (maybe it's the file opening, plus the 300ms debounce or what-not? maybe remove that specifically for file open?), but it's not very long.

*  Do you encounter any exceptions? 

Nope! And I've used it in repositories that do, and repos that do not, have smudge/clean filters. I have not checked on the other issues this PR closes.

deathaxe:
Thank you very much for your feedback. Sounds good.

You are right with the initial delay. It is caused mainly by the debounce delay which is applied to all events right now. I was already thinking about to use it for 'modified' events only, but as SublimeText sends events like 'activated', 'loaded' very quickly and several times after each other on file (re-)loads, I did not jet find a good solution for it without causing git_gutter to run too often.

I personally use `"debounce_delay": 500` to reduce latency a bit. The default is 1000ms.

kofalt:
That makes sense. I'm not afraid of thrashing I/O, so I might set that very low.

One thing worth testing is seeing what happens if the command takes longer than the debounce delay (maybe add a sleep to the command temporarily). Does a new update start processing even if an old one is still in flight? Etc.

deathaxe:
Thanks for your tip. There actually was nothing which prevented several evaluation runs at a time. 

I rebased all commits to the current master and squashed in all fixup! commits to not loose overview.

To avoid parallel runs I added a busy marker and ensure to limit debounce delay to a minimum on the other hand. On my way I decided to refactor the EventListener module, too, in order to catch the correct events.

Please give it another try. If no more issues occure, that's it.

kofalt:
Question: with your busy marker implementation, will it know to scan again after the in-flight one completes? If not, you could end up with stale data.

I'll try to give the latest version a test later today :slightly_smiling_face:

git_gutter_settings.py
r-stein: Oh sorry for the late reply. Yes it does work without a problem. I still prefer something like `return int(width) if 0 <= width <= 9 else 100000` or just `width = int(width)` a line above.

README.md
```diff
@@ -116,7 +116,17 @@ In addition the popup loads the css file `gitgutter_popup.css`, which can be mod
...
-GitGutter will show diffs in the minimap on Sublime Text 3. This can be disabled by setting `show_in_minimap` to `false`.
+
+`"show_in_minimap": 1`
+
+GitGutter shows diffs in the minimap on Sublime Text 3 by default. Change `show_in_minimap` to one of the following values to disable this feature or change the width of the markers.
+
+ value  | description
+--------|-----------------
+ 0      | hide markers
+ 1..9   | width of markers
```
r-stein: Is there any particular reason you have a fixed max width?

README.md
```diff
@@ -116,7 +116,17 @@ In addition the popup loads the css file `gitgutter_popup.css`, which can be mod
...
+ <0, >9 | show full line
```
r-stein: Better just write `-1` here

r-stein:
I think this is distracting, because it is not directly clear that the overlapping arguments belong to the `min` function and are not Region args:

``` py
region = sublime.Region(start, min(
    end, start + self._minimap_size))
```
I would prefer something like this:
``` py
region = sublime.Region(
    start, min(end, start + self._minimap_size))
```
or (if line too long):
``` py
region = sublime.Region(
    start,
    min(end, start + self._minimap_size))
```
or (if line still too long):
``` py
region_end = min(end, start + self._minimap_size)
region = sublime.Region(start, region_end)
```

deathaxe:
@kofalt: As soon as an evaluation has finished, the next received event (modified,post-saved,activated) will trigger the next run. Means as soon as you continue writing, evaluation will start again. Did not encounter any issue with this strategy right now. I may add some kind of queue to start the next evaluation as soon as the old one finished, if any event was received in between, but didn't find it needed up to now.

README.md
deathaxe: The width is calculated in number of chars. There is no technical reason for limitation, but it 9px seemed a width of gutter markers, which still looked good. 15, 20 or even 30px cover too much text, which looked somehow ugly. But this is just a personal feeling to it.

README.md
deathaxe: ok.

README.md
r-stein: I fully agree to this, but I still prefer to just allow the users to make it ugly. However you can do it as you want to :)

deathaxe: Have no problem with it. Changed!

kofalt:
Just an update: I've fallen pretty sick, so I won't be able to look at these new revisions. Might happen this weekend or something if I get better. But based on my previous experimenting it seemed to work well.

PR #376, opened 2017-02-23
Add release notes for V1.5.0

deathaxe:
I generated the changelog based on V1.4.0 with GitSavvy and did some modifications to keep only the commit messages, which seem useful.

The order is defined by the commit dates. Hope this is ok.

Do we want to create a `1.5.0-rc.1` tag? Should it include changelogs and annotation or not?

### DIFF
--- a/release_messages/src/messages/1.5.0.txt
+++ b/release_messages/src/messages/1.5.0.txt
@@ -1,0 +1,47 @@
+Changes since 1.4.0:
+
+  Enhancement:
+   - Avoid creation of temporary files, if view does not show a file from a valid git repository.
+   - Optimize number of git and diff calls.
+   - Determine, whether a view's file really exists on disk.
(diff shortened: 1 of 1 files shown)


r-stein:
I would create a pre tag since it does more than just fixing bugs.

r-stein:
For me you can merge this and #377

deathaxe:
Ok about tagging. And how about release messages? Do we want to add release messages to pre-releases, too? 

If yes, ...
1. What should be included? People who install releases only, would like to see all changes since last release, but users installing pre-releases may be interested in changes since last pre-release only. Does not make a difference here, but maybe in the future.
2. I would merge this PR and hand over the task of creating the release mesage and maybe tag to @jisaacks .

My suggestion:
- Always add changelog to TAG message
- Add release messages (x.x.x.txt) to final releases only.

PR #380, opened 2017-03-02
Other: Add announcement for pre-releases to release-message source.

deathaxe:
I added a suggestion for an anncouncement of the new pre-release strategy to the release-message
source file as suggested by FichteFoll.

Furthermore the 'Changes since ...' is replaced by the version number, so if several messages are merged, a list of versions and the changes can be displayed to the user.  Hope this is a good idea?

### DIFF
--- a/release_messages/src/messages/1.5.0.txt
+++ b/release_messages/src/messages/1.5.0.txt
@@ -1,4 +1,21 @@
-Changes since 1.4.0:
+1.5.0
+-----
+
+  ------------------------------------------------------------------------
+              !!!   WHAT HAPPEND TO GITGUTTER-EDGE   !!!
+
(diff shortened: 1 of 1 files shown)


jisaacks:
Would this be for `1.5.0` normal release? or another prerelease?

deathaxe:
I think you can use it to cut a final release. I coudn't find any issues while working with it this week and as no issues have been opened yet we are ready to publish.

r-stein:
I am also fine with it.

r-stein:
Sorry clicked the wrong button.

jisaacks:
Sweet. I'll cut a release as either tonight or this weekend!

Cheers! and HUGE thanks to you both for your continued support! 💯

deathaxe:
Added a tiny fix to master for optimization. Should be included into the release.

PR #390, opened 2017-03-20
Skip empty statuses and add commas automatically

willrowe:
This removes the need for commas at the beginning of the `status_bar_text` templates. Commas are now added automatically in between the templates.

Before, if the first template was optional, then the second one would display with a comma at the beginning even though there was nothing before it. It now checks and skips rendered templates if they are an empty string.

### DIFF
--- a/GitGutter.sublime-settings
+++ b/GitGutter.sublime-settings
@@ -82,11 +82,11 @@
   //   {{modified}} -- number of modified lines
   "status_bar_text": [
     "In {{repo}} on {{branch}}",
-    "{% if compare not in ('HEAD', branch) %}, Comparing against {{compare}}{% endif %}",
-    ", File is {{state}}",
-    "{% if deleted != 0 %}, {{deleted}}-{% endif %}",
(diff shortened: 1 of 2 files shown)


deathaxe:
The basic idea of using jinja2 engine is to enable people to **fully** customize the status bar text using the template. By moving parts of the logic to python you'll break this feature. Maybe there is someone out there, who doesn't like commas at all or wants different separators?

The logic how to render the information needs to remain in the template. This is what it's meant for.

willrowe:
Could there be a way to specify the setting as a single string instead of a list then?
It is a little confusing since it is structured as a list. I'd expect to set the entire thing in one string if I wanted to control the separator, but when setting it as a list I'd expect it to use a default separator.

Currently there is no straightforward way to have the first one optionally appear without a comma appearing at the beginning.

r-stein:
You can just use a string instead of the list. The reason for the list is that a single string would be to lange for one line and json does not support multi line strings, so we just use a list and join it. This may be not intuitive, but I don't see a better solution.

PR #394, opened 2017-03-31
Fix Compare Against when there are line breaks in git output

hedgesky:
Git output can contain `\r` symbols as part of commit message. They shouldn't be treated as separators for refs results, so use only `\n` instead.

### DIFF
--- a/git_gutter_compare.py
+++ b/git_gutter_compare.py
@@ -15,7 +15,7 @@ class GitGutterCompareCommit(object):
         def parse_results(results):
             """Parse git output and create the quick panel items."""
             if results:
-                return [r.split('\a') for r in results.splitlines()]
+                return [r.split('\a') for r in results.split('\n')]
             sublime.message_dialog('No commits found in repository.')
             return []
         return self.git_handler.git_commits().then(parse_results)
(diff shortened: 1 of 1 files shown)


deathaxe:
The list of commits and branches shows the `%s` / `%(subject)` which is by default the first line of a commit message only. So just to understand you right - where can a single line contain line breaks?

I am just interested in it as personally never saw such a situation.

git_gutter_compare.py
```diff
@@ -66,7 +66,7 @@ def parse_result(result):
         def parse_results(results):
             """Parse git output and create the quick panel items."""
             if results:
-                return [parse_result(r) for r in results.splitlines()]
+                return [parse_result(r) for r in results.split("\n")]
```
deathaxe: Use single quotes, please, to match the default quotation style used in most places / files.

deathaxe:
Please replace all `splitlines()` in the file for consistency as a further requirement to accept this PR. Thanks.

hedgesky:
Thank you for a quick response!
I've made some changes according to your requests.
About commit message: it was a Gitlab-generated merge commit with message like this:

`Merge branch 'foo' into 'develop'\r \r Update style\r \r\x07refs/heads/develop\x1asdj12398asd23`

where `\r` is a carriage return symbol. When I tried to amend this commit, the issue was gone.

PR #398, opened 2017-04-07
Add new theme: Bars Thin

callodacity:
This theme just expands on the `Bars` theme. It makes the Bars a little thinner, and it also changes the deleted markers to arrows.

**Example:**
screen shot 2017-04-08 at 07 45 04

---

I would recommend setting `line_padding_top` and `line_padding_bottom` to either `0` or `-1` in order for the bars and arrows to be connected, for example:
```
"line_padding_bottom": -1,
"line_padding_top": -1,
```


deathaxe:
Nice job.

Issue #396, opened 2017-04-02
Avoid double loading of python modules.


deathaxe:
Sublime Text actively loads all modules located in a package's root directory to search for API classes to hook. If such a module is imported by another one _(see: git_gutter.py)_ it will be loaded again by the python interpreter. All classes and globals are initialized twice, which is a waste of time and may cause issues in some cases (e.g.: settings) as multiple instances are created for each python object.

This issue can easily be avoided by moving all modules which are imported by at least one module into a folder. I already started to do move some python sources into the `modules` sub directory.

In the end all modules are imported by `git_gutter.py` need to be moved to avoid double loading.

In order not to split existing modules like git_gutter_popup.py, which may include both, the implementation and API I suggest the following folder structure.

```
[modules]
   - __init__.py    # exports API classes
   - events.py      # git_gutter_events.py
   - commands.py    # git_gutter.py
   - compare.py     # git_gutter_compare.py
   - goto.py        # git_gutter_jump_to_changes.py
   - handler.py     # git_gutter_handler.py
   - popup.py       # git_gutter_popup.py
   - settings.py    # git_gutter_settings.py
   - promise.py
   - show_diff.py   # git_gutter_show_diff.py

- plugin.py
```

The `git_gutter_` prefix of the module names was omitted as shorter is better to read and it seems unnecessary with the new structure.

This structure enables us to use relative imports on both ST2 and ST3 within the modules which makes all the `try..except..` stuff unnecessary for module imports.

The `plugin.py` simply imports the `modules` folder and `__all__` exported classes and functions. ST will see this file only, which results in the following console output upon startup.

```
reloading plugin GitGutter.plugin
```

To get an idea of how the new structure would look like, you can have a look on PR #385. It comes with the basic infrastructure required to handle this new folder/module structure.

I'd like to know if this is a structure you can agree with.

r-stein:
Have you tried some time measurement to check whether it is really faster?

I am quite neutral to it, because currently every plugin is tighten together anyway. I see as an additional advantage, that you can conditionally load plugins, e.g. load popup only in ST3. Disadvantage would be, that the files are not automatically reloaded by saving during the development. In addition it may result in some problem on updating if the function signature changed and the files are not reloaded correct.

However we should:
- pipe `plugin_loaded`/`plugin_unloaded` to the subfiles
- not abuse the `__all__` import statement, but use something else

deathaxe:
### Benchmark

[...]

PR #400, opened 2017-04-13
Restructure python modules to avoid double loading

deathaxe:
This PR intends to reduce package loading time and fix issues related by ST's double loading bug.

Related to #396

This is what this PR does.

1. Move all python sources to _modules_ folder.
2. Refactor the _git_gutter_jump_to_changes_ module.
3. Add a package reloader which once reloads modules after upgrade by Package Control


deathaxe:
Any comments?

If there is no opposition, I'd like to merge these commits to master to accomplish the new package structure and create a 1.6.0-rc.1 tag for it.

r-stein:
I am in general fine with it and think it's a good idea to improve the startup time, but I have currently not so much time+motivation to review and test it.

deathaxe:
OK, thanks. It's not a problem at all if you are low on time. I did several tests and smaller fixups during the last weeks on my own. I just want to make sure, I don't introduce changes, some of the other contributors is completely opposed to.

PR #401, opened 2017-04-14
Fix bug in protected region handling for ignored/untracked files

christek91:
While binding untracked or ignored files, new gitgutter regions would
stop being added after the loop scanning the file detected any protected
region. In effect, this means that gitgutter would only display marks on
lines before the first protected region in untracked/ignored files.


deathaxe:
Thank you very much for finding and fixing this issue.

PR #410, opened 2017-05-08
Feature: Add support information module and ISSUE_TEMPLATE.

deathaxe:
Many issues contain only vague descriptions without required information
about the environment. This makes debugging hard.

Therefore this commit adds two things to solve this problem.

1. An ISSUE_TEMPLATE should help people to make up their minds about what might be
   needed to successfully debug an issue.

2. A support module is added, which automatically collects some environment
   variables and module versions, which might be interesting for developers.

Credits
-------

This commit is inspired by the support modules from

   'BreaketHighlighter' by @facelessuser.

### DIFF
--- a/.github/ISSUE_TEMPLATE.md
+++ b/.github/ISSUE_TEMPLATE.md
@@ -1,0 +1,50 @@
+Before creating an issue, please check whether ...
+
+_1. TROUBLESHOOTING section in README can help to solve your problem._
+
+_2. an open issue exists for your problem._
+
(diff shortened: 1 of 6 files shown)


jisaacks:
I wonder if we should add a section like:

> Did you check the troubleshooting section on the README?

deathaxe:
We could and maybe this section needs some update with some of the most common issues we were facing during the past.

Especially OSX with its different environments seems a bit tricky.

Issue #411, opened 2017-05-16
When git is not installed, plus icons show in all lines


oferei:
I do not have git installed on my Windows PC. GitGutter shows plus (+) icons on all lines in all git-managed files.
Expected behavior: show some error, or do nothing at all.
Versions: GitGutter 1.6.0. Windows 10.

The console log shows many errors such as:
    GitGutter: "git rev-parse" failed with "[WinError 2] The system cannot find the file specified"
    GitGutter: "git ls-files" failed with "[WinError 2] The system cannot find the file specified"

deathaxe:
> I do not have git installed on my Windows PC. 

In this case GitGutter might not be the correct plugin for you.

> Expected behavior: show some error, or do nothing at all.

It shows the error messages compared to other packages which throw exceptions and sometimes even make ST unusable in this situation.

It is not useful to do nothing as even if git is not found it might be back on any time. You can leave ST, install git and switch back to ST. GitGutter will find git and run normally without any other interaction.

oferei:
Of course GitGutter requires git and it makes no sense to use it without git. But it should handle this error case more gracefully, communicating it better to the user: a message perhaps, not plus signs.

It happened to me because I use SourceTree which was using an embedded/walled installation of git that was not accessible from command line. I was in the process of setting up a computer, and it took me a while to figure out the problem - I did not see any error message, only the Plus signs.

deathaxe:
An error dialog is a bit tricky as GitGutter tries to run normally each time a view gets active. This is necessary to allow changed settings ("git_binary") to take effect immediately. Same with git binary. It could possibly be unavailable temporarily, which should not disable evaluation until restart of ST or requiring to reopen a file.

I am aware of this edge case and GitGutter used to act much worse in the past.

I am thinking about calling `git --version` each time a view gets activated to check whether git is working properly. I already have an version offline, which disables GitGutter in case of invalid output then. Need to finalize it.

oferei:
That's what I meant by "doing nothing". Even if displaying an error message is not possible, at least detect git availability and do not show any (plus) signs if it's missing.

I see we are aligned.

PR #412, opened 2017-05-16
Enhancement: Use git version to check if git works properly (Issue #411)

deathaxe:
Execute `git --version` to determine whether to enable GitGutterCommand.

Git version is updated each time the view is activated or a file is loaded/saved.

The git command is run synchronously without Promise() as the version needs to be available immediately in ´GitGutterCommand.is_enabled()` and there time to query version should be negligible.

The missing binary is added to a list to show error message only once per binary. As soon as git is working again, GitGutter removes it from the list and informs user about it.

### DIFF
--- a/modules/utils.py
+++ b/modules/utils.py
@@ -1,0 +1,9 @@
+# -*- coding: utf-8 -*-
+import sublime
+
+
+def log_message(msg):
+    """Print a message to statusbar and log in console."""
+    msg = 'GitGutter: {0}'.format(msg)
+    print(msg)
+    sublime.status_message(msg)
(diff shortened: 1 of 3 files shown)


modules/utils.py
```diff
@@ -0,0 +1,9 @@
+# -*- coding: utf-8 -*-
+import sublime
+
+
+def log_message(msg):
+    """Print a message to statusbar and log in console."""
+    msg = 'GitGutter: ' + msg
```
r-stein: For safety reason I would ensure that msg is a string, i.e. `msg = 'GitGutter: ' + str(msg)` or `msg = 'GitGutter: {0}'.format(msg)`

PR #414, opened 2017-05-18
Diff Popup

deathaxe:
I did a little work on diff popup module to hopefully enhance user experience with it.

### Changes

1. Prepare to keep compatible with upcoming _mdpopups 2.0.0_
2. Add support to display wrapped content/diff (requires _mdpopups 2.0.0_)
3. Remove jinja2 template handling of CSS files as it is done by _mdpopups_ already
4. Optical enhancements of the popup itself
   a) fix border width
   b) adapt colors to color scheme
   c) improve visibility on light color schemes
   d) add some spacing around the navigation buttons
   e) navigation buttons try to use fonts which are known to render the symbols well
   f) use background colors to highlight diffs and distinguish replacements/inserts/deletions 
5. Avoid flickering if highlighting diff is enabled/disabled
6. Use the algorithm of difflib.Differ to create the highlighted diff as it produces better diffs in most cases

### Notes

1. All these changes made heavy changes to the _gitgutter_popup.css_ necessary, so users might update their local definitions if used. 
2. Diff Popup requires Sublime Text 3119+ and mdpopups 1.9.0+ needed for wrapper classes and CSS variable support.
3. _mdpopups 2.0.0_ will require Sublime Text 3124
4. TODO: Update README

### Internals
1. Some changes to adapt code style of the other modules
2. Split module into several parts to keep file sizes low.

### DIFF
--- a/gitgutter_popup.css
+++ b/gitgutter_popup.css
@@ -41,9 +41,26 @@
  **/
 html.dark {
     background-color: color(var(--background) blend(#fff 90%));
+    --diffpopup-ins-bg: color(green blend(black 60%));
+    --diffpopup-ins-fg: color(green blend(white 40%));
+    --diffpopup-del-bg: color(red blend(black 60%));
+    --diffpopup-del-fg: color(red blend(white 40%));
+    --diffpopup-chg-ins-bg: color(#769908 blend(black 95%));
+    --diffpopup-chg-ins-fg: color(#769908 blend(white 20%));
+    --diffpopup-chg-del-bg: color(orangered blend(black 80%));
+    --diffpopup-chg-del-fg: color(orangered blend(white 40%));
 }
+
 html.light {
     background-color: color(var(--background) blend(#000 90%));
+    --diffpopup-ins-bg: color(green blend(white 25%));
+    --diffpopup-ins-fg: color(green blend(black 80%));
+    --diffpopup-del-bg: color(red blend(white 25%));
+    --diffpopup-del-fg: color(red blend(black 80%));
+    --diffpopup-chg-ins-bg: color(#93921b blend(white 20%));
+    --diffpopup-chg-ins-fg: color(#93921b blend(black 85%)); /*#b3ca23*/
+    --diffpopup-chg-del-bg: color(orangered blend(white 20%));
+    --diffpopup-chg-del-fg: color(orangered blend(black 85%));
 }
 
 /* Compatibility rules for correct popup borders with mdpopups 1.x.x */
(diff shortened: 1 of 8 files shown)


modules/popup/factory.py
```diff
@@ -0,0 +1,314 @@
...
+
+            active icon button: <a href="revert"><symbol>⟲</symbol></a>
+            active text button: <a href="revert"><text>(revert)</text></a>
+            inactive icon button: <symbol>⟲</symbol>
+            inactive text button: <text>(revert)</text>
+    """
+    # The format to render disabled/enabled buttons
+    button_format = ('<{0}>{1}</{0}>', '<a href="{2}"><{0}>{1}</{0}></a>')
+    # The tag to use for each button
+    button_tag = ('text', 'symbol')
```
r-stein: I think we should just remove the "text" caption. It is just way to big for the popup and icon issues have been fixed in the newest versions.

gitgutter_popup.css
```diff
@@ -1,49 +1,187 @@
-/*
-    Jinja2 css template. Provided values are:
-        - st_version: the version of Sublime Text;
+/**********************************************************************
```
r-stein: I think this css is not support for 3118 and before. Not sure how problematic that is.

modules/popup/factory.py
deathaxe: I agree with you in general and never used text, too. Already thought about it, too but was not sure, whether there are people who prefer text. Therefore I kept it.

gitgutter_popup.css
deathaxe: mdpopups 2.0.0 which is planed to be released in June, will require Sublime Text 3124 anyway.

As the wrapper class is obligatory since GitGutter 1.5.0 and supported by mdpopups only if ST 3119+ is installed, we might have issues with earlier builts anyway.

I will add an announcement to the next release notes / messages about requirements.

ST 3119 is about 1 year old now and it is the first built with nearly complete minihtml support, so I guess it is acceptable to make it a requirement.

modules/popup/factory.py
deathaxe: > icon issues have been fixed in the newest versions.

But is not in 3119+ which will still be supported.

r-stein: Yes, but 
1. even in 3119 it is not so common 
2. with the "text buttons" the button line is way to large and not better than just having the "missing symbol" icons and remember the order
3. I don't think people really dig into the settings and decided to use it (people who do something like this also update their ST)
4. I started with text buttons and moved to symbols and in the hindsight I should have removed them in the initial PR

deathaxe: I have no problem to remove it at all. I personally agree with you in all points. Makes code simpler.

jisaacks:
FYI, @r-stein understands the popup much better than I do, so if he approves, than I approve.

PR #415, opened 2017-05-26
Release 1.7.0

deathaxe:
I think the number of bugfixes and the amount of rework of the diff popup are worth a release.

If nobody is opposed, I would ask @jisaacks to make the cut.


jisaacks:
Cool. As long as theres no opposition I'll cut this either tonight or tomorrow when I get time. :)

PR #445, opened 2017-09-14
Internal: Refactor built and release system

deathaxe:
GitGutter's built and release system relies on ruby, make and bash scripts.

1. They are not available by nature on every operating system.
2. Sublime Text's python engine can do the same thing without external dependencies.
3. The release message files exist twice in the repository.

The goals of this commit are to ...

1. Built & publish releases directly from ST3.
2. Store release messages in "messages" folder once as all the other packages do.
3. Keep the opportunity to make releases from the shell if desired.

---

With this commit, releasing a new version of GitGutter requires the following steps:

1. Set-Up the GitHub access token (needed once)

   - Call "Preferences: GitGutter Settings" from command pallet
   - Add a key "github_token": "..." to the user file.

2. Create a proper message file in the "messages" folder.

   - The name is the version being used to release.
   - The first line of the file contains the version as heading for ST.
   - The global message is not required (it is added by step 3)

3. Call "GitGutter: Build Release" from command pallet

   This will create a "Cut 1.x.x" commit and tag locally.

4. Call "GitGutter: Publish Release" from command pallet

   This will ...
   - push the latest commit of master branch
   - push the release tag
   - create a release with the release message from the most recent messages/1.x.x.txt

The "Release" commands are visible only, if the "github_token" is set up.

---

The release.py also provides a CLI to run it from shell as follows:

  ./release build
  ./release publish --token <thetoken>

The makefile was modified as well.

---

All release-related files are added to the .gitattributes to prevent them from being deployed in ordinary sublime-package files.


deathaxe:
Yes I assumed it constant. Therefore I added the template hard-coded to the release.py

Do you plan changes?

jisaacks:
> Do you plan changes?

I don't currently have any changes planned, but it's changed before. I don't foresee any changes anytime soon, and if there are in the future, it will just be one occurrence that doesn't get stripped so I don't think it's that bad.

PR #447, opened 2017-09-15
Enhancement: Use shellenv library to help finding git on Linux/MacOS.

deathaxe:
This PR adds an optional import of 'shellenv' library to read environment variables from a user's login shell to be used for git execution.

As a result GitGutter always runs the git binary also seen by the login shell. This change should hopefully help preventing issues on Linux or especially OSX with git binary not being found or wrong version being used.

GitGutter should work without https://github.com/int3h/SublimeFixMacPath now.

Can someone with a Mac check, whether it really works and if it is useful to add this commit to GitGutter?

### DIFF
--- a/dependencies.json
+++ b/dependencies.json
@@ -2,14 +2,16 @@
     "*": {
         "*": [
             "markupsafe",
-            "python-jinja2"
+            "python-jinja2",
+            "shellenv"
         ],
         ">=3119": [
             "pygments",
             "python-markdown",
             "mdpopups",
             "python-jinja2",
-            "markupsafe"
+            "markupsafe",
+            "shellenv"
         ]
     }
 }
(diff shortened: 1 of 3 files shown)


deathaxe:
Seems latest ST versions publish the login shell's environment to python already.

See: https://forum.sublimetext.com/t/how-do-i-get-sublime-to-look-in-usr-local-bin/24141

A setting was added to Preferences to manipulate this behavior:

```
    // OS X only. If the user's default shell should be invoked to obtain the
    // user's customized environment variables. May be a boolean, or a string
    // of the path the shell to invoke. Sublime Text must be restarted for this
    // to take effect.
    "shell_environment": true,
```

With that in mind it might be redundant to add another dependency to GitGutter which does the same thing as _SublimeFixMacPath_ does a well job on older setups already.

Thoughts?

Issue #476, opened 2017-11-28
On-gutter-hover popups conflict


braver:
We ran into this when developing a popup for SublimeLinter. When hovering the gutter, it should show a popup with all linter errors for a given line. However, in combination with GitGutter, the GitGutter popup tends to show instead. It kinda feels like it’s a race condition that decides which shows. 

There is this protected region setting, and that does help deciding which icon shows in the gutter. And that works. Perhaps it should also be used to decide if a popup should be shown. As far as I can tell, protected regions aren’t checked for popups. Which is good for bookmarks etc. 

So I don’t know what a good approach would be to improve the interop between these two plugins, but I do want to put this out there in hope someone has a brilliant idea ☺️

deathaxe:
Basically I can see three options:

1. Use the existing `protected_regions`
2. Add a new `popup_protected_regions` to disable diff popup for regions known to provide higher priority popup functionality.
3. Show diff popup only for lines occupied by a `git_gutter_...` region.

Option 1. and 3. should result in equal behavior while 3. is faster and easier to check as all `git_gutter_...` regions start at the beginning of a line so their beginning boundary is already equal to the `point` argument of the `on_hover` handler. It would further result in a consistent behavior as diff popup might not be expected if no gutter mark exists. But as you already stated, it would disable diff popup even though there is no other popup to show which is true for bookmarks and maybe others.

Option 2 would add more flexibility to decide when to show the diff popup. With a little bit of help from foreign packages it could even be much faster than option 1 or 3. What I have in mind is each popup (e.g. SublimeLinter) adding an invisible `sublime_linter.popup_region` if it shows a popup, with the `point` from `on_hover`. The popup would need to implement the `on_hide` event to remove the region again.

Example:

```
import sublime
import sublime_plugin

class EventListener(sublime_plugin.EventListener):

    def on_hover(self, view, point, hover_zone):

        failed = False

        if hover_zone != sublime.HOVER_GUTTER:
            return

        def remove_region():
            """Remove popup region, if popup gets hidden."""
            view.erase_regions("package_name.popup_region")

        def on_async_done():
            """An async task is completed with the popup being finally added."""
            if failed:
                return remove_region()

            view.show_popup(content="Hello World",
                            location=point,
                            on_hide=remove_region)

        # simulate async task
        sublime.set_timeout_async(on_async_done, 200)

        # mark line as occupied by "package_name"
        view.add_regions("package_name.popup_region",
                        regions=sublime.Region(point),
[...]

PR #477, opened 2017-11-29
Enhancement: Add protected regions for diff popup (Issue #476)

deathaxe:
This PR adds support to ignore the hover event if a line is occupied by a certain region. The diff popup is not displayed by hovering such lines, but still can be opened by key binding or command pallet.

This enhancement is meant to improve compatibility with other packages.

### DIFF
--- a/GitGutter.sublime-settings
+++ b/GitGutter.sublime-settings
@@ -137,6 +137,12 @@
         "lsp_info"
     ],
 
+    // Do not show the Diff Popup if a line contains these regions
+    // Useful for making sure GitGutter does not fight with other popups.
+    "diff_popup_protected_regions": [
+        "sublime_linter.protected_regions"
+    ],
+
     // The gutter theme defines the icons to show for the different events.
     "theme": "Default.gitgutter-theme"
 }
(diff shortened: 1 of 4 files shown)


r-stein:
I don't think this in general solves the problem with other packages, but is a solution especially focused on SublimeLinter. I would prefer a more general solution, which also scales to other packages and is not only for GitGutter (I think if GitGutter and SublimeLinter start with general solution other developers will join), but I have no idea how this solution could look like.

We may wait few days and see of someone has a clever idea and merge it if no one has a suggestion.

deathaxe:
It provides an option to disable diff popup when hovering the gutter area with the already existing technology of regions. It is somewhat stupid as it may need user interaction to work around popup fighting. The solution targets popup fighting with a certain hover area only, that's true.

The only general solution was to register all "show_popup" calls by name and create a global priority list which is used by ST to decide which popup to show if more than one is requested to be displayed. 

Therefore the only general solution was to improve popups management functionality in ST itself. Same applies to gutter icons, which would need some kind of priority, too.

r-stein:
I agree that this may be "good enough" for real world use cases =)

PR #480, opened 2018-01-23
settings diff algorithm getter fix

b-fett:
### Package settings:
image
### Before patch:
image
### After patch:
image


deathaxe:
Good catch. This is indeed the wrong setting we try to read.

PR #489, opened 2018-03-05
Fix README documentation for `diff_algorithm`

ErinCall:
The actual default is `default`, so the readme shouldn't claim it's `patience`. It's already documented in the description table, so there's no need to mention it twice.

Also fixed a typo.


deathaxe:
Thanks

PR #496, opened 2018-05-01
Fix: Compatibility issue of GitGutter with Snap on Ubuntu

### LINKED ISSUE #495: Snap not working on Ubuntu 18.04
# Support Info

- Sublime Text 3143
- Platform: linux
- Arch: x64
- GitGutter 1.7.14
- Install via PC: True
- git version 2.17.0
- mdpopups 3.3.3
- markdown 2.6.11
- pygments 2.1a0
- jinja2 2.8

## Steps to Reproduce Issue

1. Install GitGutter via snap on Ubuntu 18.04
2. Try to use it

```
GitGutter failed to create git cache: [Errno 2] No such file or directory: '/run/user/1000/snap.sublime-text/GitGutter'
Exception AttributeError: AttributeError("'TempFile' object has no attribute 'name'",) in <bound method TempFile.__del__ of <GitGutter.modules.temp.TempFile object at 0x7facfd3cbb10>> ignored
GitGutter failed to create view cache: [Errno 2] No such file or directory: '/run/user/1000/snap.sublime-text/GitGutter'
Exception AttributeError: AttributeError("'TempFile' object has no attribute 'name'",) in <bound method TempFile.__del__ of <GitGutter.modules.temp.TempFile object at 0x7facfd3cb890>> ignored
Package Control: Skipping automatic upgrade, last run at 2018-05-01 10:20:56, next run at 2018-05-01 11:20:56 or after
```

deathaxe:
Fixes #495 

If Sublime Text is installed via snap package manager on Ubuntu the XDG_RUNTIME_DIR environment variable points to a directory like '/run/user/<pid>/snap.sublime-text/'.

The 'snap.sublime-text' does not exist. Therefore GitGutter fails creating its 'GitGutter' sub directory.

This commit makes use of pythons `os.makedirs()` to ensure the whole path tree is created if required.

Furthermore the `name` and `_mode` attributes are moved to the top of `__init__()` to avoid `AttributeError`s in `__del__()`, if temp-file creation fails.

### DIFF
--- a/modules/temp.py
+++ b/modules/temp.py
@@ -52,26 +52,25 @@ class TempFile(object):
 
     def __init__(self, mode='r'):
         """Initialize TempFile object."""
+        self.name = tempfile.mktemp(dir=TEMP_DIR)
         self._file = None
+        self._mode = mode
         # Cache unlink to keep it available even though the 'os' module is
         # already None'd out whenever __del__() is called.
         # See python stdlib's tempfile.py for details.
         self._unlink = os.unlink
 
         try:
-            os.mkdir(TEMP_DIR)
+            # ensure cache directory exists with write permissions
+            os.makedirs(TEMP_DIR, 0o555)
         except OSError as e:
-            # Need to use OSError as FileExistsError is not supported by ST2!
             if e.errno != errno.EEXIST:
                 raise
 
-        self.name = tempfile.mktemp(dir=TEMP_DIR)
-        self._mode = mode
-
     def __del__(self):
         """Destroy the TempFile object and remove the file from disk."""
-        self.close()
         try:
+            self.close()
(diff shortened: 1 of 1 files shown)


modules/temp.py
```diff
@@ -52,26 +52,25 @@ class TempFile(object):
 
     def __init__(self, mode='r'):
         """Initialize TempFile object."""
+        self.name = tempfile.mktemp(dir=TEMP_DIR)
```
r-stein: Doesn't it result in a an issue to create the temp file before creating the directory?

deathaxe: `mktemp` does not touch the filesystem in any way. It just creates the path. Using `mkstemp()`  would cause an issue.

PR #526, opened 2018-11-02
Add option to ignore whitespace changes when blaming in line annotations

sciyoshi:
Adds a new setting `line_annotation_ignore_whitespace` which controls passing `-w` to `git blame`. This is great if you have a repo where you want to ignore a commit has changed a bunch of whitespace (see [Git's docs about the option](https://git-scm.com/docs/git-blame#git-blame--w)).

### DIFF
--- a/GitGutter.sublime-settings
+++ b/GitGutter.sublime-settings
@@ -154,6 +154,14 @@
     //   ...
     "line_annotation_ruler": false,
 
+    // (ST3, Build 3124+ only)
+    // Whether to ignore whitespace changes when showing line annotations.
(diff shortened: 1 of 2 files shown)


deathaxe:
Thanks for your effort.

I've pulled your changes and did some cosmetic changes:

1. possible values in sublime-settings should be double quoted hence PackageDev uses them for auto-completion.
2. Added the new setting to the _Preferences.sublime-settings-hints_ so PackageDev marks it a valid setting in _Preferences_
3. Add an entry to the documentation.

Issue #528, opened 2018-11-07
IndexError thrown every time ST starts


rwols:
## Description

Every time I open sublime text, I see this exception in the console:

```
Traceback (most recent call last):
  File "/Applications/Sublime Text.app/Contents/MacOS/sublime_plugin.py", line 462, in run_callback
    expr()
  File "/Applications/Sublime Text.app/Contents/MacOS/sublime_plugin.py", line 586, in <lambda>
    run_callback('on_selection_modified', callback, lambda: callback.on_selection_modified(v))
  File "/Users/raoulwols/Library/Application Support/Sublime Text 3/Installed Packages/GitGutter.sublime-package/modules/events.py", line 278, in on_selection_modified
    self._run_blame(view, False)
  File "/Users/raoulwols/Library/Application Support/Sublime Text 3/Installed Packages/GitGutter.sublime-package/modules/events.py", line 289, in _run_blame
    row, _ = view.rowcol(selection[0].begin())
  File "/Applications/Sublime Text.app/Contents/MacOS/sublime.py", line 649, in __getitem__
    raise IndexError()
IndexError
```

## Support Info

- Sublime Text 3180
- Platform: osx
- Arch: x64
- GitGutter 1.9.4
- Install via PC: True
- git version 2.17.2 (Apple Git-113)
- mdpopups 3.3.3
- markdown 2.6.11
- pygments 2.1a0
- jinja2 2.8

## Steps to Reproduce Issue

1. Close ST.
2. Open ST.
3. Open console.

deathaxe:
Can you checkout #529 please and try whether it fixes your issue and does not introduce any new side effects?

deathaxe:
Fixed in 1.9.5

PR #529, opened 2018-11-08
Enhancement: line annotation display events

deathaxe:
This commit makes some changes to the event handler, which is responsible to
call git blame and display line annotations.

The origin for the changes is issue #528. Even though the issue could not be
reproduced on Windows a couple possible enhancements were identified.

Fixes

1. Fail silently if the active row can't be identified because of an invalid
   or empty selection object returned by ST API.

Enhancements:

1. Hide line annotation if...
   a) multiple cursors are active
   b) selection is not empty
   c) caret is moved vertically in row

   As a result line annotation is displayed only if a single cursor is moved
   to a new unmodified line of text, if word wrap is disabled.

   This should help improving the editing experience and reduce the impact of
   some of the phantoms' drawbacks / limitations.

2. All checks are performed within the debounced code to avoid the line
   annotation to flicker up if text is selected via mouse.

### DIFF
--- a/modules/events.py
+++ b/modules/events.py
@@ -280,30 +280,34 @@ class BlameEventListener(sublime_plugin.EventListener):
     def _run_blame(self, view, force):
         """Run blame if the caret moves to another row."""
 
-        # check if caret is present
-        selection = view.sel()
-        if not selection:
-            return
-
-        # do nothing if row didn't change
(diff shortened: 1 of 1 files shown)


rwols:
Hi @deathaxe, I just checked out this branch and it makes the exception disappear. Nice work!

PR #548, opened 2019-07-03
Fix property naming not adjusted

Pictor13:
Thank you for the quick fix @deathaxe .
There was an error in the console with a variable not defined, since some code was still referencing the property named as public (`template`) rather than private (`_template`).

This should fix it.
If I missed anything else to adjust, let me know in the comments.

### DIFF
--- a/modules/statusbar.py
+++ b/modules/statusbar.py
@@ -117,8 +117,8 @@ class GitGutterStatusBar(object):
     def is_enabled(self):
         """Return whether status bar text is enabled in settings or not."""
         enabled = self.settings.get('show_status_bar_text', False)
-        if self.template and not enabled:
-            self.template = None
+        if self._template and not enabled:
+            self._template = None
             self.vars['repo'] = None
             self.erase()
         return enabled
@@ -136,7 +136,7 @@ class GitGutterStatusBar(object):
                 False - if no variable is used by the template.
         """
         try:
-            return any(var in self.template.variables for var in variables)
+            return any(var in self._template.variables for var in variables)
         except:
             return False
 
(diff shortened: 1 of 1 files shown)


modules/statusbar.py
```diff
@@ -117,8 +117,8 @@ def template(self):
     def is_enabled(self):
         """Return whether status bar text is enabled in settings or not."""
         enabled = self.settings.get('show_status_bar_text', False)
-        if self.template and not enabled:
-            self.template = None
+        if self._template and not enabled:
+            self._template = None
```
deathaxe: Good point

modules/statusbar.py
```diff
@@ -136,7 +136,7 @@ def has(self, variables):
                 False - if no variable is used by the template.
         """
         try:
-            return any(var in self.template.variables for var in variables)
+            return any(var in self._template.variables for var in variables)
```
deathaxe: We need to keep the `self.template` here in order to create the template lazily on demand.

modules/statusbar.py
```diff
@@ -162,4 +162,4 @@ def update(self, **kwargs):
 
         if want_update:
             self.view.set_status(
-                '00_git_gutter', self.template.render(**self.vars))
+                '00_git_gutter', self._template.render(**self.vars))
```
deathaxe: Same here: We need to keep the `self.template` here in order to create the template lazily on demand.

deathaxe:
Fixed the commented lines, squashed and merged commits.

Pictor13:
Great!
Will the plugin update itself via Package Control or should I do a manual update?

Btw, sorry for the incorrect fix; I didn't pay enough attention and renamed quickly (and blindly 🙈).
Rushed; almost no confidence with Python; plus disattention... 💥🙄

deathaxe:
This is what pull requests are - let have another two eyes a look onto a change.

As you confirm it to work I can cut a release for Package Control users.

PR #565, opened 2021-06-22
typo

aiqc:
changed "an" to "a"


deathaxe:
Thanks! Tweaked the fix a bit.

Issue #563, opened 2021-05-31
ST4 - config "show_line_annotation" not working


jocke76:
Hi

After upgrade to Sublime Text 4 (4107) GitGutter shows 'blame info' after each line, even though 'show_line_annotation' is false.
(Mac OS 10.14.6 Mojave)

deathaxe:
Maybe unhappy documented (see #553).

Did you really set `"show_line_annotation": false`?

The comment shows `"false"` with quotes so PackageDev can pick it up for completions. It's still meant to be entered as `false` though.

jocke76:
Hi deathaxe

Sorry for being unclear. 
This is my actual configuration:

<img width="511" alt="Screenshot 2021-06-01 at 07 38 16" src="https://user-images.githubusercontent.com/6542217/120271840-591a6e80-c2ac-11eb-8a42-903f84ee6820.png">


And below is an example from the code.
<img width="1765" alt="Screenshot 2021-06-01 at 07 36 42" src="https://user-images.githubusercontent.com/6542217/120271746-253f4900-c2ac-11eb-8243-a92697401674.png">

jocke76:
Update!

Oh, now I understand what you meant =)
I changed it from "false" to false, and now it works!
Pre ST4 it has been working with "false" though, so I guess ST3 was more forgiving, or?

deathaxe:
> Pre ST4 it has been working with "false" though, so I guess ST3 was more forgiving, or?

No, this issue has been reported for ST3 last time. Nothing changed with regards to how ST parses settings. A `git_gutter_show_line_annotation` in Preferences.sublime-settings whould override the setting from GitGutter.sublime-settings.

Basically any setting can be placed into Preferences if prefixed with `git_gutter_`. This way project specific settings are powered.

I should probably add some code to support string format, too now that it is documented as such.

PR #570, opened 2022-11-19
Fixed incorrectly quoted boolean hints

brian6932:
:)


deathaxe:
They are quoted by intent in order for PackageDev to pick values up for completions.

see: #563

brian6932:
Ah ok, it was a bit confusing when I saw it

